PR #1, opened 2020-11-20
add aggregate tests

DavidBadura:


Makefile
```diff
@@ -0,0 +1,31 @@
+help:                                                                           ## shows this help
+	@awk 'BEGIN {FS = ":.*?## "} /^[a-zA-Z_\-\.]+:.*?## / {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST)
+
+.PHONY: php-cs-check
+php-cs-check:                                                                   ## run cs fixer (dry-run)
+	PHP_CS_FIXER_FUTURE_MODE=1 vendor/bin/php-cs-fixer fix --allow-risky=yes --diff --dry-run
```
DanielBadura, line 6: risky is part of the config, we should remove this here

Makefile
```diff
@@ -0,0 +1,31 @@
...
+
+.PHONY: php-cs-fix
+php-cs-fix:                                                                     ## run cs fixer
+	PHP_CS_FIXER_FUTURE_MODE=1 vendor/bin/php-cs-fixer fix --allow-risky=yes
```
DanielBadura, line 10: same as above

composer.json
```diff
@@ -27,6 +27,7 @@
     "friendsofphp/php-cs-fixer": "^2.16.7",
     "phpstan/phpstan": "^0.12.54",
     "phpunit/phpunit": "^9.4.3",
+    "symfony/var-dumper": "^5.1",
```
DanielBadura, line 30: Do we really need this?

DanielBadura (approved):
I think the patch is fine as it is, only some minor things

composer.json
DavidBadura: As a development tool, I would leave it in because I use it a lot.

PR #7, opened 2020-12-07
improve aggregate root

DavidBadura:


src/Repository/Repository.php
```diff
@@ -108,12 +107,12 @@ private function assertExtendsEventSourcedAggregateRoot(string $class): void
      */
     private function createAggregate(array $eventStream): AggregateRoot
     {
-        $reflectionClass = new ReflectionClass($this->aggregateClass);
+        $class = $this->aggregateClass;
 
-        /** @var AggregateRoot $aggregate */
-        $aggregate = $reflectionClass->newInstanceWithoutConstructor();
-        $aggregate->initializeState($eventStream);
+        if (is_subclass_of($class, AggregateRoot::class) === false) {
```
DanielBadura, line 112: this should not be necessary since we are already asserting this in the creation of the class. Maybe we just need to harden the type on the property?

DanielBadura: if yes, the whole method could be ditched i think

DavidBadura: i moved this check to the constructor and improved the type internally.

DanielBadura (approved):

PR #10, opened 2020-12-08
dependencies in Makefiles so composer deps are up to date

tolry:
> untested :stuck_out_tongue: 

This will automatically run `composer install` when executing any task that depends on composer dependencies being up to date. It does that by comparing the timestamps of the vendor folder with those of the files `composer.lock` and `composer.json`.

Have made this change 'blindly' with github web-editor, but I am using that quite succesfully e.g. here https://github.com/tolry/bcr/blob/master/Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,31 +1,35 @@
 help:                                                                           ## shows this help
 	@awk 'BEGIN {FS = ":.*?## "} /^[a-zA-Z_\-\.]+:.*?## / {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST)
 
+vendor: composer.json composer.lock
+	composer install
+	
 .PHONY: php-cs-check
-php-cs-check:                                                                   ## run cs fixer (dry-run)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,31 +1,35 @@
 help:                                                                           ## shows this help
 	@awk 'BEGIN {FS = ":.*?## "} /^[a-zA-Z_\-\.]+:.*?## / {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST)
 
+vendor: composer.json composer.lock
```
DanielBadura, line 4: wouldn't it be sufficient to just look at the lock file?

DanielBadura (approved):

Makefile
tolry: yeah, probably, since a change to composer.json without composer.lock would need a `composer update` anyways, feel free to adjust as needed

DavidBadura:
tanks @tolry  :)

PR #13, opened 2021-01-08
change php-cs-fixer to phpcs

DavidBadura:


phpcs.xml.dist
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0"?>
+<ruleset name="patchlevel/enum coding standard">
```
DanielBadura, line 2:
```suggestion
<ruleset name="patchlevel/event-sourcing coding standard">
```

src/Aggregate/AggregateChanged.php
```diff
@@ -75,8 +81,8 @@ public function recordNow(int $playhead): self
      */
     public static function deserialize(array $data): self
     {
-        /** @var string $class */
         $class = $data['event'];
+        assert(is_string($class));
```
DanielBadura, line 85: could we use webmozart/assert ?

DavidBadura: the phpcs fixer did that, I have now solved it with an exception. i don't want to add a library for one check.

DavidBadura:
I have exclude SuperfluousSuffix for now, we can refactor the exceptions in another PR.

DanielBadura (approved):

PR #15, opened 2021-01-08
add snapshot concept

### LINKED ISSUE #22: add snapshots

DavidBadura:
close #22


src/Snapshot/SnapshotStore.php
```diff
@@ -0,0 +1,17 @@
...
+use Patchlevel\EventSourcing\Aggregate\SnapshotableAggregateRoot;
+
+interface SnapshotStore
+{
+    public function save(Snapshot $snapshot): void;
+
+    /**
+     * @param class-string<SnapshotableAggregateRoot> $aggregate
+     */
+    public function load(string $aggregate, string $id): ?Snapshot;
```
DanielBadura, line 16: i would be more happy if the api would not contain a `null` case and instead throw an exception

src/Aggregate/AggregateRoot.php
```diff
@@ -13,7 +13,7 @@ abstract class AggregateRoot
 {
     /** @var AggregateChanged[] */
     private array $uncommittedEvents = [];
-    private int $playhead = -1;
+    protected int $playhead = -1;
```
DanielBadura, line 16: im not quite happy with this but i dont think we can let it be as private since we need to read, set & update the value in the child class.

DanielBadura (changes requested):
UnitTest are missing, these should be added with the feature

src/Aggregate/AggregateRoot.php
DavidBadura: at least i set it to internal

src/Snapshot/SnapshotStore.php
DavidBadura: done

src/Aggregate/AggregateRoot.php
DavidBadura: i don't think we should start at 0 with the playhead. i find 1 more logical at the first event.

DavidBadura: we should discuss that in another issue

DanielBadura (approved):

PR #18, opened 2021-01-08
fix deserialize

DavidBadura:
Adjusted the test to show a deserialization problem. In this PR it will be fixed later.

### DIFF
--- a/src/Store/SingleTableStore.php
+++ b/src/Store/SingleTableStore.php
@@ -4,9 +4,12 @@ declare(strict_types=1);
 
 namespace Patchlevel\EventSourcing\Store;
 
+use DateTimeImmutable;
 use Doctrine\DBAL\Connection;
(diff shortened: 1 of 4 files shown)


src/Store/SingleTableStore.php
```diff
@@ -211,4 +213,26 @@ private static function shortName(string $name): string
...
+        if (!$result['recordedOn']) {
+            return $result;
+        }
+
+        $recordedOn = Type::getType(Types::DATETIMETZ_IMMUTABLE)->convertToPHPValue(
+            $result['recordedOn'],
+            $platform
+        );
+
+        assert($recordedOn instanceof DateTimeImmutable);
```
DanielBadura, line 233: i really dislike `assert`. An if checks that throws or `webmozart/assert` is in my opinion more suitable.

DanielBadura (approved):

PR #24, opened 2021-01-14
add aggregate hashmap in store for shortname and tablename

DavidBadura:


src/Store/MultiTableStore.php
```diff
@@ -101,7 +107,11 @@ public function has(string $aggregate, string $id): bool
      */
     public function saveBatch(string $aggregate, string $id, array $events): void
     {
-        $tableName = self::tableName($aggregate);
+        if (!array_key_exists($aggregate, $this->aggregates)) {
+            throw new AggregateNotDefined($aggregate);
+        }
+
+        $tableName = $this->aggregates[$aggregate];
```
DanielBadura, line 114: maybe extract into private method `getTableName(): string`

DanielBadura (approved):

PR #31, opened 2021-01-28
Provide the ability to change the table name of the SingleTableStore

### LINKED ISSUE #26: allow to change the table name in SingleTableStore

DanielBadura:
fixes #26

### DIFF
--- a/src/Store/SingleTableStore.php
+++ b/src/Store/SingleTableStore.php
@@ -23,12 +23,12 @@ final class SingleTableStore extends DoctrineStore
     /**
      * @param array<class-string<AggregateRoot>, string> $aggregates
      */
-    public function __construct(Connection $connection, array $aggregates)
+    public function __construct(Connection $connection, array $aggregates, string $tableName)
     {
         parent::__construct($connection);
 
         $this->aggregates = $aggregates;
-        $this->tableName = 'eventstore';
+        $this->tableName = $tableName;
     }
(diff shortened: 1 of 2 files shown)


src/Store/SingleTableStore.php
```diff
@@ -29,6 +28,7 @@ public function __construct(Connection $connection, array $aggregates)
         parent::__construct($connection);
 
         $this->aggregates = $aggregates;
+        $this->tableName = 'eventstore';
```
DavidBadura, line 31: why didn't you add it via the constructor signature?

src/Store/SingleTableStore.php
```diff
@@ -162,9 +163,14 @@ public function schema(): Schema
         return $schema;
     }
 
+    public function changeTableName(string $tableName): void
```
DavidBadura, line 166: I would not change it afterwards, that could destroy things.

src/Store/SingleTableStore.php
DanielBadura: the plan was to have the same signature like before & multi table store, but you are right with the comment below. The damage could be devastating if someone change the name of the table in a running system. So i guess its way better to provide the value via ctor but i would then drop the default name since defaulting parameters always tend to be a maintenance burden, wdyt?

DavidBadura (approved):

PR #35, opened 2021-01-29
pipeline & commands

### LINKED ISSUE #27: research/concept: how to rebuild projections
It must be possible to build a projection from scratch. To do this, you have to get the necessary events from the DB. The difficulties here are that a projection can consist of several aggregates and therefore different tables have to be queried. In addition, the amount of data should not be neglected.

DavidBadura:
fix #27


src/Store/StreamableStore.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Patchlevel\EventSourcing\Store;
+
+use Generator;
+use Patchlevel\EventSourcing\Aggregate\AggregateChanged;
+
+interface StreamableStore
```
DanielBadura, line 10: shouldnt it extend the store interface?

DavidBadura:
the first idea. i have no idea how it should work with multitable. so that the projection can be rebuilt, we need the events in the correct order.

with multitable i just have to load all events from all tables and merge/sort them via php. which can eat up a lot of memory. you can also combine tables on the database side with union, but dbal does not support this directly. you would have to write native sql statements.

src/Console/ProjectionRebuildCommand.php
```diff
@@ -0,0 +1,63 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+class ProjectionRebuildCommand extends Command
+{
+    private Store $store;
+    private ProjectionRepository $projectionRepository;
+
+    public function __construct(Store $store, ProjectionRepository $projectionRepository)
```
DanielBadura, line 23: Since we are only supporting `StreamableStore` we could typhint it right here instead of duck type checking it at the execution

src/Pipeline/Source/StoreSource.php
```diff
@@ -0,0 +1,32 @@
...
+    {
+        $this->store = $store;
+    }
+
+    /**
+     * @return Generator<EventBucket>
+     */
+    public function load(): Generator
+    {
+        return $this->store->all();
```
DanielBadura, line 25: we could use here `yield from`

src/Console/ProjectionRebuildCommand.php
DanielBadura: lets investigate this again some time later

DanielBadura (approved):

PR #44, opened 2021-02-02
improve commands

DavidBadura:


Makefile
```diff
@@ -24,6 +24,10 @@ psalm: vendor
 phpunit: vendor                                                                 ## run phpunit tests
 	vendor/bin/phpunit --testdox --colors=always -v $(OPTIONS)
 
+.PHONY: psalm
```
DanielBadura, line 27:
```suggestion
.PHONY: infection
```

phpunit.xml.dist
```diff
@@ -20,6 +20,7 @@
     </coverage>
 
     <php>
+        <env name="COLUMNS" value="120"/>
```
DanielBadura, line 23: do we still need this?

DavidBadura: yes, we need this. otherwise we get line breaks in the middle of the sentence.

DanielBadura (approved):

PR #46, opened 2021-02-03
Harden types

### LINKED ISSUE #43: Try to harden the types even more

DanielBadura:
fixes #43

### DIFF
--- a/src/Store/DoctrineStore.php
+++ b/src/Store/DoctrineStore.php
@@ -12,6 +12,8 @@ use Doctrine\DBAL\Types\Type;
 use Doctrine\DBAL\Types\Types;
 use Patchlevel\EventSourcing\Aggregate\AggregateChanged;
 
+use function is_int;
+
 abstract class DoctrineStore implements Store
 {
     protected Connection $connection;
--- a/src/Store/DoctrineStore.php
+++ b/src/Store/DoctrineStore.php
@@ -29,23 +29,37 @@ abstract class DoctrineStore implements Store
     abstract public function schema(): Schema;
 
     /**
-     * @param array{aggregateId: string, playhead: ?int, event: class-string<AggregateChanged>, payload: string, recordedOn: string} $result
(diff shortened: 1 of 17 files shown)


src/Store/MultiTableStore.php
```diff
@@ -39,7 +40,7 @@ public function __construct(
     /**
      * @param class-string<AggregateRoot> $aggregate
      *
-     * @return AggregateChanged[]
+     * @return array<AggregateChanged>
```
DavidBadura, line 43:
```suggestion
     * @return list<AggregateChanged>
```

src/Store/DoctrineStore.php
```diff
@@ -28,27 +29,37 @@ public function connection(): Connection
...
         if (!$recordedOn instanceof DateTimeImmutable) {
             throw new StoreException('recordedOn should be a DateTimeImmutable object');
         }
 
-        $result['recordedOn'] = $recordedOn;
+        return $recordedOn;
+    }
 
-        return $result;
+    private static function normalizePlayhead(string $playheadAsString, AbstractPlatform $platform): int
```
DavidBadura, line 55: simple int cast should be fine here

src/Store/MultiTableStore.php
```diff
@@ -96,7 +97,7 @@ public function has(string $aggregate, string $id): bool
 
     /**
      * @param class-string<AggregateRoot> $aggregate
-     * @param AggregateChanged[]          $events
+     * @param array<AggregateChanged>     $events
```
DavidBadura, line 100:
```suggestion
     * @param list<AggregateChanged> $events
```

src/Store/MultiTableStore.php
DanielBadura: cant be done unless using `array_values` after `array_merge`. Seems not feasible for this

src/Store/MultiTableStore.php
DanielBadura: cant be done unless using `array_values` after `array_merge`. Seems not feasible for this

PR #47, opened 2021-02-03
Add ShowCommand to view all Events for an aggregate, Add WatchCommand that prints all incoming events into CLI

DavidBadura:


src/WatchServer/WatchServerClient.php
```diff
@@ -0,0 +1,128 @@
...
+    {
+        if (strpos($host, '://') === false) {
+            $host = 'tcp://' . $host;
+        }
+
+        $this->host = $host;
+        $this->socket = null;
+    }
+
+    public function send(AggregateChanged $event): bool
```
DanielBadura, line 43: Couldnt we get the singature to void instead of boolean and throw different exception for the different cases where the send was not successful instead only returning false?

DanielBadura (approved):

src/WatchServer/WatchServerClient.php
DavidBadura: done

PR #56, opened 2021-04-12
add event bucket index

DavidBadura:


DanielBadura:
Which usecase do we have rot this index based approach?

DavidBadura:
I'm thinking of how to get projection asynchronous. Or rebuild the projection separately. And so that everything works, in my opinion, you have to know at which event you stopped in order to continue there.

DavidBadura:
@DanielBadura This is important so that we can run parts of the stream in the future. Such as a cron job that delays the processing of events and always remembers where it has stopped. Or a async runner, who always pulled the last events.

DanielBadura (approved):

PR #57, opened 2021-04-12
add until option in rebuild command

DavidBadura:


src/Pipeline/Pipeline.php
```diff
@@ -36,13 +36,13 @@ public function run(?callable $observer = null): void
...
         foreach ($this->source->load() as $bucket) {
+            $observer($bucket);
+
             $result = ($this->middlewares)($bucket);
 
             foreach ($result as $resultBucket) {
                 $this->target->save($resultBucket);
             }
-
-            $observer($bucket);
```
DanielBadura, line 45: Do you know why this was executed after the middleware seems this was wrong from the beginning doesnt it ?

DavidBadura: I guess I didn't think about it. I only noticed it while editing.

src/Console/Command/ProjectionRebuildCommand.php
```diff
@@ -34,7 +40,8 @@ protected function configure(): void
         $this
             ->setName('event-sourcing:projection:rebuild')
             ->setDescription('rebuild projection')
-            ->addOption('recreate', 'r', InputOption::VALUE_NONE, 'drop and create projections');
+            ->addOption('recreate', 'r', InputOption::VALUE_NONE, 'drop and create projections')
+            ->addOption('until', 'u', InputOption::VALUE_OPTIONAL, 'create the projection up to a point in time [2017-02-02 12:00]');
```
DanielBadura, line 44: Value should be required if option is given

src/Pipeline/Pipeline.php
DanielBadura: okey 👍

DanielBadura (approved):

PR #62, opened 2021-07-01
Add psalm template for payload to extend it as an user

DanielBadura:


DanielBadura:
Should be fixed by "accident" by #61

PR #81, opened 2021-11-23
Create .gitattributes to prevent unneeded files will be exported on publish to packagist

DanielBadura:


DanielBadura:
CI failed because of #75 and the not runned workflows..

PR #85, opened 2021-11-29
add database commands

DavidBadura:


src/Console/Command/DatabaseDropCommand.php
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->setName('event-sourcing:database:drop')
+            ->setDescription('create eventstore database');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
```
DanielBadura, line 32: maybe we should here add a defensive mechanism, so it will only warn if normal executed and only execute if `--force` is added or sth. like that?

Im not sure about this just throwing this idea here out

DavidBadura: that was first a test so that I could try it out in my project. I have expanded it now ;-)

PR #92, opened 2021-11-30
rename apply and handle method

DavidBadura:
* apply -> record
* handle -> apply


DanielBadura:
ping @MarcoGiesen @bartrail
this will be a bc break

DanielBadura (approved):

PR #94, opened 2021-12-01
add strict apply method & extract the default apply method

DavidBadura:


DanielBadura:
Not quite sure if we should go this route. Need some time to think about it. The patch itself looks so far good.

DanielBadura (approved):

PR #95, opened 2021-12-01
add clock

DavidBadura:
I think it would be a good idea to add a clock implementation to this library. We can use it to write our tests better and the users can also manipulate the time for their tests. This is the missing part for the "testing" documentation.

### DIFF
--- a/src/Aggregate/AggregateChanged.php
+++ b/src/Aggregate/AggregateChanged.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Patchlevel\EventSourcing\Aggregate;
 
 use DateTimeImmutable;
-use Patchlevel\EventSourcing\Clock;
 
 use function json_decode;
 use function json_encode;
@@ -123,6 +122,6 @@ abstract class AggregateChanged
 
     protected function createRecordDate(): DateTimeImmutable
     {
-        return Clock::createDateTimeImmutable();
+        return new DateTimeImmutable();
     }
 }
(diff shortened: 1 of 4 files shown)


DanielBadura:
I dont think we should do this. The user has already the possibility to replace the method `createRecordDate` with are "more suitable" one if desired. And manipulating `DateTime` is always tricky and really painful if done wrong. I think with https://github.com/patchlevel/event-sourcing/pull/58 we already did the job. We dont **really** need this for our testing usecase, do we?

DavidBadura:
Yes, you can override it, but it is not good for the dx if the user need to create an own "mock" or "class" to manage clock features. Other event sourcing libraries do it too: https://eventsauce.io/docs/utilities/clock/

DanielBadura:
Hm, i can see the need for it generally but the solution used by eventsauce is somewhat cleaner from my PoV since it's a different class used for mocking and normal usage. With our solution here it could happen that in production the clock is frozen. 

But i cannot provide another solution to this tbh since we cannot inject different classes here in this situation. That's why I'm concered about the consequeces for some dx could be a fatal failure for production. I know, with the eventsauce approach this could also happen, but it seems for me that this is more unlikely.

Maybe a path between would be suitable? Just add the clock class but not directly use it in the `AggregateChanged` directly?

src/Aggregate/AggregateChanged.php
```diff
@@ -5,6 +5,7 @@
 namespace Patchlevel\EventSourcing\Aggregate;
 
 use DateTimeImmutable;
+use Patchlevel\EventSourcing\Clock;
```
tolry, line 8: NIH? :slightly_smiling_face: https://github.com/lcobucci/clock

DavidBadura: We cannot do any service injection at this point. And features like sleep, so that you skip a certain period of time, are also not available.

tolry: :disappointed:

DanielBadura (approved):

PR #96, opened 2021-12-01
add schema provider for migration

DavidBadura:


DanielBadura:
I'm not quite sure for which case we need it here. We already have mirgations support with the bundle, do we?

DavidBadura:
Doctrine Migration has nothing to do with symfony or bundle. It is glue code for doctrine migration and can be used for other frameworks like laminas too.

DanielBadura:
Ah yes, im fine with that. Didnt think about that right away.

DanielBadura (approved):

PR #100, opened 2021-12-03
cleanup exceptions

### LINKED ISSUE #93: standardize exceptions
some exceptions are in subfolders, some has a exceptions suffix and some don't have it. We should standardize it.

DavidBadura:
fix #93

### DIFF
--- a/src/Repository/WrongAggregate.php
+++ b/src/Repository/WrongAggregate.php
@@ -8,8 +8,14 @@ use function sprintf;
 
 final class WrongAggregate extends RepositoryException
 {
+    /**
+     * @param class-string $aggregateClass
+     * @param class-string $expected
+     */
     public function __construct(string $aggregateClass, string $expected)
     {
-        parent::__construct(sprintf('aggregate "%s::%s" not found', $aggregateClass, $expected));
+        parent::__construct(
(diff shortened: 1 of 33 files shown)


src/Repository/WrongAggregate.php
```diff
@@ -4,11 +4,9 @@
...
 namespace Patchlevel\EventSourcing\Repository;
 
-use RuntimeException;
-
 use function sprintf;
 
-final class WrongAggregateException extends RuntimeException
+final class WrongAggregate extends RepositoryException
 {
     public function __construct(string $aggregateClass, string $expected)
```
DanielBadura, line 11: We should update the message here too, cannot add a suggestion since the line wasnt change:
`Wrong aggregate given: got '%s' but expected '%s'`

src/Store/SingleTableStore.php
```diff
@@ -95,7 +95,7 @@ public function has(string $aggregate, string $id): bool
         );
 
         if (!is_int($result) && !is_string($result)) {
-            throw new StoreException('invalid query return type');
+            throw new InvalidType('invalid query return type');
```
DanielBadura, line 98: This exception with the exact same message could be a dedicated one since it is used frequently

DavidBadura: how would you call them?

DanielBadura (approved):

PR #106, opened 2021-12-05
refactor aggregate changed for better psalm supprt

DavidBadura:


phpcs.xml.dist
```diff
@@ -13,6 +13,10 @@
 
     <rule ref="PatchlevelCodingStandard">
         <exclude name="SlevomatCodingStandard.Classes.SuperfluousExceptionNaming.SuperfluousSuffix" />
+        <exclude name="SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingTraversableTypeHintSpecification" />
+        <exclude name="SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingTraversableTypeHintSpecification" />
+        <exclude name="SlevomatCodingStandard.TypeHints.PropertyTypeHint.MissingTraversableTypeHintSpecification" />
+        <exclude name="SlevomatCodingStandard.Commenting.InlineDocCommentDeclaration.InvalidFormat" />
```
DanielBadura, line 19: Could we only disable them in place instead of globally?

DavidBadura: The MissingTraversableTypeHintSpecification should be only checked by psalm and should not be needed in phpcs.

And if we want multi-line doc comments, we should remove the InlineDocCommentDeclaration .InvalidFormat too.

DanielBadura: Are these inline multiline doc even "legal"? maybe we should revert this part.

As of the TypeHintSpecification rules: if the would really go this route, then we would need to remove alot of rules in the coding-standard. I personally have no objections at phpcs also yelling at us if it can find something that is not right typewise. The question for me would be where this happens and is it a false positive which we could report upstream or is it really a fail from us.

DavidBadura: > Are these inline multiline doc even "legal"? maybe we should revert this part.

i don't know, we can revert this.

> if the would really go this route, then we would need to remove a lot of rules in the coding-standard.

I would only remove the rules if they are produce false positives. We have now a lot of such errors because phpcs don't understand generics and template annotations. For me, it make not sense, that phpcs check if a type is right because psalm and phpstan can do it better.

DanielBadura (approved):

.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea
```
DavidBadura:
```suggestion
```

Issue #134, opened 2021-12-18
add php 8.1 and drop 7.4 support


DavidBadura:

robinlehrmann:
Can be closed 🎉

DavidBadura:
done in #135

PR #135, opened 2021-12-18
removed php7.4, added support php8.1

robinlehrmann:
belongs to https://github.com/patchlevel/event-sourcing/issues/134

### DIFF
--- a/.github/workflows/coding-standard.yml
+++ b/.github/workflows/coding-standard.yml
@@ -19,7 +19,7 @@ jobs:
                 dependencies:
                     - "locked"
                 php-version:
-                    - "7.4"
+                    - "8.0"
                 operating-system:
                     - "ubuntu-latest"
 
--- a/.github/workflows/mutation-tests.yml
+++ b/.github/workflows/mutation-tests.yml
@@ -19,7 +19,7 @@ jobs:
                 dependencies:
                     - "locked"
                 php-version:
-                    - "7.4"
+                    - "8.0"
                 operating-system:
                     - "ubuntu-latest"
 
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -19,7 +19,8 @@ jobs:
                 dependencies:
                     - "locked"
                 php-version:
-                    - "7.4"
+                    - "8.0"
+                    - "8.1"
                 operating-system:
                     - "ubuntu-latest"
 
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -20,7 +20,6 @@ jobs:
                     - "locked"
                 php-version:
                     - "8.0"
-                    - "8.1"
                 operating-system:
                     - "ubuntu-latest"
 
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -21,8 +21,8 @@ jobs:
                     - "highest"
                     - "locked"
                 php-version:
-                    - "7.4"
                     - "8.0"
+                    - "8.1"
                 operating-system:
                     - "ubuntu-latest"
                     - "windows-latest"
--- a/.github/workflows/psalm.yml
+++ b/.github/workflows/psalm.yml
@@ -19,7 +19,8 @@ jobs:
                 dependencies:
                     - "locked"
                 php-version:
-                    - "7.4"
+                    - "8.0"
+                    - "8.1"
                 operating-system:
                     - "ubuntu-latest"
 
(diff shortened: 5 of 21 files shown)


DavidBadura:
you need to change the php versions in the `.github/workflow`  too ;-)

.github/workflows/coding-standard.yml
```diff
@@ -19,7 +19,8 @@ jobs:
                 dependencies:
                     - "locked"
                 php-version:
-                    - "7.4"
+                    - "8.0"
+                    - "8.1"
```
DavidBadura: static code analysis should only run on the lowest php version.

composer.json
```diff
@@ -19,14 +19,13 @@
     }
   ],
   "require": {
-    "php": "^7.4|^8.0",
-    "ext-json": "^7.4|^8.0",
+    "php": "^8.0|^8.1",
```
DavidBadura: `^8.0` includes `8.1`

.github/workflows/phpunit.yml
```diff
@@ -21,7 +21,6 @@ jobs:
                     - "highest"
                     - "locked"
                 php-version:
-                    - "7.4"
```
DavidBadura, line 24: 8.1 is missing

DavidBadura (changes requested):

.github/workflows/mutation-tests.yml
```diff
@@ -19,7 +19,8 @@ jobs:
                 dependencies:
                     - "locked"
                 php-version:
-                    - "7.4"
+                    - "8.0"
+                    - "8.1"
```
DavidBadura: here too, pls run mutation checks only on the lowest php version

robinlehrmann:
@DavidBadura
Do you know why this is failing ? https://github.com/patchlevel/event-sourcing/runs/4569869039?check_suite_focus=true 😅
In this line is nothing 🤔 `vendor/symfony/console/Helper/Table.php:564`
I also didn't find a `strstr` in your `ShowCommandTest` 😄

DavidBadura:
The error was here: https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Console/Helper/Table.php#L566

symfony [fixed](https://github.com/symfony/symfony/pull/41157) it in a patch already. I bump the minimum version in this pr #136. If you rebase your branch it should be fixed now :)

.github/workflows/phpstan.yml
```diff
@@ -19,7 +19,8 @@ jobs:
                 dependencies:
                     - "locked"
                 php-version:
-                    - "7.4"
+                    - "8.0"
+                    - "8.1"
```
DavidBadura: Same here too, only the lowest version has to be tested.

.github/workflows/psalm.yml
```diff
@@ -19,7 +19,8 @@ jobs:
                 dependencies:
                     - "locked"
                 php-version:
-                    - "7.4"
+                    - "8.0"
+                    - "8.1"
```
DavidBadura: Same here too, only the lowest version has to be tested.

DavidBadura (changes requested):

src/Store/DoctrineStore.php
```diff
@@ -55,7 +55,7 @@ private static function normalizeRecordedOn(string $recordedOnAsString, Abstract
         return $recordedOn;
     }
 
-    private static function normalizePlayhead(string $playheadAsString, AbstractPlatform $platform): int
+    private static function normalizePlayhead(string|int $playheadAsString, AbstractPlatform $platform): int
```
DanielBadura, line 58: If we open up the type here we should also do it here: https://github.com/patchlevel/event-sourcing/pull/135/files#diff-17ca8e151b14a68bb5bf016108d5793f3d6bb69ca630897acaf44f2f8c4993c5R33 since it can be `int` as it seems. 

This would mean at this points we should also adjust it:
* https://github.com/patchlevel/event-sourcing/blob/ba3818fef58f7959255281d90e74aabd69d598de/src/Store/SingleTableStore.php#L53
* https://github.com/patchlevel/event-sourcing/blob/ba3818fef58f7959255281d90e74aabd69d598de/src/Store/MultiTableStore.php#L57

robinlehrmann: @DanielBadura I changed it with: `playhead: string|int`

.github/workflows/phpstan.yml
robinlehrmann: ups.. yes, done :)

DavidBadura (approved):

DavidBadura (approved):

DanielBadura (approved):

DanielBadura:
Thanks @robinlehrmann !

PR #144, opened 2021-12-22
set required php version to fixed minor versions

DavidBadura:


DavidBadura:
@DanielBadura do you know how i can skip some matrix combination for the phpunit workflow?

DanielBadura:
Yes, we are already doing it with the dependency entry: https://github.com/patchlevel/event-sourcing/blob/1.1.x/.github/workflows/phpunit.yml#L60

DavidBadura:
There we skip only one "step". i want to skip the whole job.

DanielBadura (approved):

PR #154, opened 2021-12-28
add backward-compatibility-check workflow

### LINKED ISSUE #105: Introduce  Roave/BackwardCompatibilityCheck as CI job

### LINKED ISSUE #111: Add Roave/BackwardCompatibilityCheck as CI Job
https://github.com/Roave/BackwardCompatibilityCheck

DavidBadura:
fix #105


DanielBadura:
I think the should push out the tools like this into a separate `composer.json` since alot of deps are getting pulled in which the user does not need. The SA tools (psalm, phpstan, phpcs) could stay in the normal json i think. WDYT @DavidBadura ?

DavidBadura:
Yes, let's do it that way, we can simply adjust it again if it doesn't fit :)

.github/workflows/backward-compatibility-check.yml
```diff
@@ -0,0 +1,50 @@
...
+        with:
+          dependency-versions: ${{ matrix.dependencies }}
+
+      - uses: "ramsey/composer-install@v2"
+        with:
+          dependency-versions: ${{ matrix.dependencies }}
+          working-directory: 'tools'
+
+      - name: "Check for BC breaks"
+        run: "tools/vendor/bin/roave-backward-compatibility-check --format=github-actions --install-development-dependencies"
```
DanielBadura, line 50: not quite sure if we need the dev deps here..

DavidBadura:
LGTM

PR #157, opened 2022-01-01
fix table schema example in projection documentation

robinlehrmann:
Added missing name column in projection example


DavidBadura:
@robinlehrmann thanks :)

PR #158, opened 2022-01-02
add attribute based apply methods

DavidBadura:


tests/Unit/Aggregate/AggregateRootWithAttributeApplyTest.php
```diff
@@ -0,0 +1,82 @@
...
+        $id = ProfileId::fromString('1');
+        $email = Email::fromString('david.badura@patchlevel.de');
+
+        $target = ProfileId::fromString('2');
+
+        $profile = ProfileWithAttributeApply::createProfile($id, $email);
+        $profile->visitProfile($target);
+
+        self::assertEquals('1', $profile->aggregateRootId());
+        self::assertEquals(1, $profile->playhead());
```
robinlehrmann: why not `self::assertSame` for scalar values?

docs/aggregate.md
```diff
@@ -295,9 +295,99 @@ all newly recorded events are then fetched and written to the database.
...
+final class Profile extends AggregateRoot
+{
+    use AttributeApplyMethod;
+
+    private string $id;
+    private string $name;
+
+    // ...
+    
+    #[Apply(ProfileCreated::class)]
```
robinlehrmann, line 324: `#[Apply]`

What do you think to make the `aggregateChangedClassName` optional and use the first parameter type as aggregateChangedClassName? 
See blow

src/Aggregate/AttributeApplyMethod.php
```diff
@@ -0,0 +1,73 @@
...
+        $methods = $reflector->getMethods();
+
+        $map = [];
+
+        foreach ($methods as $method) {
+            $attributes = $method->getAttributes(Apply::class);
+
+            foreach ($attributes as $attribute) {
+                $instance = $attribute->newInstance();
+
```
robinlehrmann: just as an idea maybe :)
```
$aggregateChangedClass = $instance->aggregateChangedClass();
if (!$aggregateChangedClass) {
    $parameters = $method->getParameters();

    if (count($parameters) !== 1) {
        // throw logic custom exception
    }

    $aggregateChangedClass = (string) $parameters[0]->getType();
}

$map[$aggregateChangedClass] = $method->getName();
```

DavidBadura: Yeah, i think about it too, but this is not easy. You have union `Foo|Bar` and Intersection `Foo&Bar` types, so you need cover this too. And in the future, maybe you can write this `Foo|(Bar&Baz)`, so we need to cover this cases too or throw an exception. Make this optional is not a BC break and we can check this in the future, if it make sense. But for now, i think we should say it explicit.

tests/Unit/Aggregate/AggregateRootWithAttributeApplyTest.php
DanielBadura: We have already typesafety since the methods are declaring `string` or `int` as return value. But sure, `assertSame` should be always preferred i think.

DavidBadura: we can do this with help of reactor or php-cs-fixer over the whole codebase.
https://github.com/patchlevel/event-sourcing/issues/160

docs/aggregate.md
```diff
@@ -295,12 +295,145 @@ all newly recorded events are then fetched and written to the database.
 The `apply` method must be implemented so that the events are processed on the aggregate. 
 This method can get quite big with some events.
 
-To make things structured, you can use two different traits 
+To make things structured, you can use three different traits 
 that allow you to define different apply methods for each event.
 
+### Attribute based apply method (since v1.2)
+
+The first variant is the preferred variant.
```
DanielBadura:
```suggestion
The first variant is the preferred one. 
```

phpstan.neon.dist
```diff
@@ -5,3 +5,4 @@ parameters:
 	level: max
 	paths:
 		- src
+	checkGenericClassInNonGenericObjectType: false
```
DanielBadura, line 8: tbh I'm more a fan of adding them to the baseline then to deactivated the whole feature for two reasons.

1. This deactivates all checks, not only there where we cant fix it.
2. The baseline can cleanup itself, if we fix it, or phpstan evolves and fixes it itself, since it can understand what we are doing

src/Attribute/Suppress.php
```diff
@@ -0,0 +1,54 @@
...
+namespace Patchlevel\EventSourcing\Attribute;
+
+use Attribute;
+use InvalidArgumentException;
+use Patchlevel\EventSourcing\Aggregate\AggregateChanged;
+
+use function is_string;
+
+#[Attribute(Attribute::TARGET_CLASS)]
+class Suppress
```
DanielBadura: Im not sure if i like the idea itself but if the do this i think the naming can be improved:
`SuppressMissingApplyMethods` would be more explaining i think

phpstan.neon.dist
DavidBadura: This rule only leads to false positives for us and has not yet caught anything that was correct. We have psalm what Generics understands correctly and has implemented better.

And it's not just a problem that we always have when we do something. But everyone who wants to help develop this library.

DanielBadura: Well we could add a contribution guideline where we can state out that psalm is for this project first citizen analyser and phpstan not but i get what you mean.

DavidBadura: I commented on phpstan and see what the feedback is:

https://github.com/phpstan/phpstan/issues/6240#issuecomment-1007364430

DanielBadura: Hm let's see what the result will be. But until then we would need a decision for this PR.

I would be fine with merging it as it is, since psalm is here our preferred SA tool.

DanielBadura (approved):

PR #161, opened 2022-01-07
add handle attribute for projections

DavidBadura:


tests/Unit/Projection/DefaultProjectionRepositoryTest.php
```diff
@@ -128,6 +131,77 @@ public function drop(): void
...
+            Email::fromString('profile@test.com')
+        );
+
+        $projectionRepository = new DefaultProjectionRepository([$projection]);
+        $projectionRepository->handle($profileCreated);
+
+        self::assertSame($profileCreated, $projection::$handledEvent);
+
+        $profileVisited = ProfileVisited::raise(ProfileId::fromString('1'), ProfileId::fromString('2'));
+        $projectionRepository->handle($profileVisited);
```
DanielBadura: The assert if the method got handled is missing

src/Projection/AttributeHandleMethod.php
```diff
@@ -0,0 +1,57 @@
...
+use Patchlevel\EventSourcing\Aggregate\AggregateChanged;
+use Patchlevel\EventSourcing\Attribute\Handle;
+use ReflectionClass;
+
+use function array_key_exists;
+
+/**
+ * @psalm-require-implements Projection
+ */
+trait AttributeHandleMethod
```
DanielBadura, line 16: we have a lot of traits atm tbh 😨

DavidBadura: :man_shrugging:

DavidBadura: Do you have a better idea? I think that in the next major release we can force attribute support and thereby remove all traits. But would still wait and collect ideas.

tests/Unit/Projection/DefaultProjectionRepositoryTest.php
DavidBadura: i rewrite the tests :)

src/Projection/AttributeHandleMethod.php
DanielBadura: No, no other idea atm

DanielBadura (approved):

PR #163, opened 2022-01-11
Update analysers (pslam, phpstan & infection)

DanielBadura:
Somehow dependabot got a bit lazy...

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -6388,16 +6388,16 @@
         },
         {
             "name": "vimeo/psalm",
-            "version": "v4.15.0",
+            "version": "4.18.1",
             "source": {
                 "type": "git",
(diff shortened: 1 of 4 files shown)


composer.lock
```diff
@@ -4891,46 +4891,42 @@
         },
         {
             "name": "symfony/console",
-            "version": "v5.4.1",
+            "version": "v6.0.2",
```
DanielBadura, line 4894: This here is triggering the BC warning. Im not quite sure but i would say this is no bc for us. If we would say it's a BC then we could not support 3 Major Versions. If we would go the route to only support 1 major version, it would be fine from my site and then this would be a real bc break.

DavidBadura (approved):

PR #166, opened 2022-01-12
fix case insensitive column names

### LINKED ISSUE #165: postgres case insensitive column names
Postgres columns always seem to be created in lowercase. Saving the data works without any problems. But when reading you get the column names in lowercase.

Example:

`recordedOn` => `recordedon`

There is then an error when accessing the array. Another field is also affected: `aggregateId`.

The solution would be to create the columns in `lowersnakecase`. But that would be a **BC break** because you would have to adapt the scheme.

Otherwise you would have to fix the data in the store so that the data can be loaded from Postgres.

DavidBadura:
fix #165 

and added integration tests for:

* sqlite
* postgres: 9.4, 13, 14
* mysql: 5.7, 8.0
* mariadb: 10.0, 10.2, 10.5


src/Store/DoctrineStore.php
```diff
@@ -55,7 +63,10 @@ private static function normalizeRecordedOn(string $recordedOnAsString, Abstract
         return $recordedOn;
     }
 
-    private static function normalizePlayhead(string $playheadAsString, AbstractPlatform $platform): int
+    /**
+     * @param string|int $playheadAsString
```
DanielBadura, line 67: we should adjust the type for `normalizeResult` parameter too then.

DanielBadura (approved):

PR #182, opened 2022-03-02
change schema fields to snake case

### LINKED ISSUE #170: change schema to snake case
Some databases do not support camelcase columns (postgres). We should always create our schema in snakecase. Also the tables.

DavidBadura:
fix #170


DanielBadura (approved):
LGTM

PR #185, opened 2022-03-04
split message and event

### LINKED ISSUE #172: split message and domain events
We currently mix domain events and message information with `AggregateChanged` class. That should be resolved.

We should remove meta/header information from the `AggregateChanged` (maybe rename it to DomainEvent):

* aggregate id
* playhead
* recorded on

Instead, move that information into a "Message" class, which then wraps the domain event. This happens automatically when reading out the events in the aggregate root.

In addition, the aggregate class-name can also be added to the message, so that the EventBucket (which is a form of Message) can be replaced in the pipeline. In other words, the pipeline then uses the same messages.

The store will then only work with messages and not even with AggregateChanged and once with EventBucket.

The EventBus is also converted to messages. So also the processors and projection.

DavidBadura:
resolve #172


DanielBadura (approved):
The only thing that is bothering me is that we loose one more "feature" and that is the recordedOn date. The user has with this patch no possibilty to change it. But tbh, i think this is good regulary the only thing is testing: there they need to use our Clock class. I think it would be better if the Clock class could be changed for testing but i doubt it is possible.

DavidBadura:
Actually, developers shouldn't need to know anything more about the fact that this recordedOn date exists and that it is just meta information that they can't use directly. 

Previously we had the problem that it was difficult to compare events generated by aggregates with expected events since this meta information is recordedOn and playhead part of the event. Now it is no longer part of the event. In other words, if you want information like createdAt etc, you have to add it to the event yourself. And so the developer has it in his own hands. I don't even know when the developer would want to test entire messages where it might be relevant.

PR #186, opened 2022-03-04
refactor projections

DavidBadura:
* switch to attributes (for create & drop too)
* rename ProjectionRepository to ProjectionHandler
* allow create, drop and recreate specific projection over cli


src/Projection/ProjectionMetadata.php
```diff
@@ -0,0 +1,20 @@
...
+declare(strict_types=1);
+
+namespace Patchlevel\EventSourcing\Projection;
+
+use Patchlevel\EventSourcing\Aggregate\AggregateChanged;
+
+/**
+ * @internal
+ */
+class ProjectionMetadata
```
DanielBadura: final maybe if this is internal?

src/Projection/DefaultProjectionHandler.php
```diff
@@ -0,0 +1,130 @@
...
+
+            if ($method->getAttributes(Create::class)) {
+                $metadata->createMethod = $method->getName();
+            }
+
+            if (!$method->getAttributes(Drop::class)) {
+                continue;
+            }
+
+            $metadata->dropMethod = $method->getName();
```
DanielBadura: If adding the Attribute to more then 1 Method in the class then the last one would be taken, i would suggest to throw an exception here, since it makes literally no sense to provide multiple create or drop methods. This only would lead to wtf moments so we should be here really explicit imho.

DanielBadura (commented):
I somewhat like and dislike these changes. I extremely like the handle part in the repositories. But dislike the Pipeline changes and this weird API with the `DefaultProjectionHandler`.

For Drop & Create i'm neutral. I'm a fan of expilicit api's and this is going to wash it out a bit.

src/Console/Command/ProjectionRebuildCommand.php
```diff
@@ -23,17 +22,15 @@
...
 use function sprintf;
 
-final class ProjectionRebuildCommand extends Command
+final class ProjectionRebuildCommand extends ProjectionCommand
 {
     private Store $store;
-    private ProjectionRepository $projectionRepository;
 
-    public function __construct(Store $store, ProjectionRepository $projectionRepository)
+    public function __construct(Store $store, ProjectionHandler $projectionRepository)
```
DanielBadura: $projectionHandler

DanielBadura (approved):

PR #193, opened 2022-03-19
hydrator

### LINKED ISSUE #188: event serializer
It would be interesting if we could support new PHP 8.1 constructs.

```php
class ProfileCreated {
    public function __construct(
        private readonly string $name,
        #[Transformer(EmailTransformer::class)]
        private readonly Email $email,
    ) {}
}
```

DavidBadura:
fix #188


src/Serializer/DefaultHydrator.php
```diff
@@ -0,0 +1,136 @@
...
+     * @param array<string, mixed> $data
+     *
+     * @return T
+     *
+     * @template T of object
+     */
+    public function hydrate(string $class, array $data): object
+    {
+        $object = self::reflectionClass($class)->newInstanceWithoutConstructor();
+        assert($object instanceof $class);
```
DanielBadura, line 34: would say: lets take an if + exception here instead of assert

src/Serializer/DefaultHydrator.php
```diff
@@ -0,0 +1,136 @@
...
+            $normalizer = null;
+            $attributeReflectionList = $property->getAttributes(Normalize::class);
+
+            if ($attributeReflectionList !== []) {
+                $attribute = $attributeReflectionList[0]->newInstance();
+                $normalizer = $attribute->normalizer();
+            }
+
+            if (!is_string($property->name)) {
+                throw new RuntimeException('how?');
```
DanielBadura: Better message? :D

src/Serializer/DefaultHydrator.php
```diff
@@ -0,0 +1,136 @@
...
+        $reflectionClass = self::reflectionClass($class);
+        $metadata = [];
+
+        foreach ($reflectionClass->getProperties() as $property) {
+            $property->setAccessible(true);
+
+            $normalizer = null;
+            $attributeReflectionList = $property->getAttributes(Normalize::class);
+
+            if ($attributeReflectionList !== []) {
```
DanielBadura, line 97: count should be faster i think

DanielBadura (approved):

src/Serializer/DefaultHydrator.php
DavidBadura: This is a case that can never happen. It's only for the psalm, because otherwise he wouldn't understand it. I don't really want to do a check, so assert fits better since it's ignored in production.

src/Serializer/DefaultHydrator.php
DavidBadura: I really don't know if it's faster or slower. From a feeling I would say that my version should be faster because it doesn't "count" here. Even if opcache optimization happens here, the optimization should also be applicable to my variant. I have searched for it but haven't really found anything. Are you 100% sure it is faster?

PR #201, opened 2022-03-25
use php 8.1 for tools

DavidBadura:


baseline.xml
```diff
@@ -9,6 +9,11 @@
       <code>$value</code>
     </MixedAssignment>
   </file>
+  <file src="src/Store/MultiTableStore.php">
+    <UnnecessaryVarAnnotation occurrences="1">
```
DanielBadura, line 13: this could be removed, if you remove the `@var` annotation here https://github.com/patchlevel/event-sourcing/pull/201/files#diff-c9f691442fd7c0a3f2d16d8a9e7ee619b51308a1a8b3b5f31822e206b0313025R176

DavidBadura: yes, but then i get 15 errors in phpstan. I think this time it's better to have one error in psalm baseline than 15 errors in phpstan.

DanielBadura (approved):

PR #202, opened 2022-03-25
Event metadata

### LINKED ISSUE #174: event name mapper
Maybe we should introduce an event name mapper. The full-class name is currently used to store the event name in the db.But it would be good if there was an opportunity to store the event name in the database without the class name. You would need a hashmap variant for that.

DavidBadura:
closes #174 
closes #192


src/Serializer/JsonSerializer.php
```diff
@@ -5,48 +5,173 @@
...
+    /** @var array<class-string, string> */
+    private array $eventClassMapRevert;
+
+    /** @var array<class-string, ReflectionClass> */
+    private array $reflectionClassCache = [];
+
+    /**
+     * @param array<string, class-string> $eventClassMap
+     */
+    public function __construct(EventMetadataFactory $metadataFactory, array $eventClassMap = [])
```
DanielBadura: I think we should add a EventRegistry class for the events since checking if a class is registered or not should not be the serializers concern. So the code below would look like this:
```php
$eventName = $this->eventRegistry->getByClassName($event::class);
```

instead of 
```php
       if (!array_key_exists($event::class, $this->eventClassMapRevert)) {
            throw new EventClassNotRegistered($event::class);
        }

        $eventName = $this->eventClassMapRevert[$event::class];
```

And also adding `$eventRegistry->getByEventName(): string`

Wdyt?

src/Serializer/JsonSerializer.php
```diff
@@ -5,48 +5,173 @@
...
     {
-        $this->hydrator = $hydrator ?? new DefaultHydrator();
+        $this->metadataFactory = $metadataFactory;
+        $this->eventClassMap = $eventClassMap;
+        $this->eventClassMapRevert = array_flip($eventClassMap);
     }
 
-    public function serialize(object $event): string
+    public function serialize(object $event, array $options = []): SerializedData
```
DanielBadura, line 37: i think we should add here the keys we are using as an doc-type for the options array. So `array{pretty_print: bool}`

src/Serializer/Serializer.php
```diff
@@ -6,14 +6,15 @@
 
 interface Serializer
 {
-    public function serialize(object $event): string;
+    public const OPTION_PRETTY_PRINT = 'pretty_print';
```
DanielBadura, line 9: i think this should move to our implementation since this has no impact on the API for the implementing side.

src/Serializer/JsonSerializer.php
```diff
@@ -5,48 +5,173 @@
...
+
+        assert($object instanceof $class);
+
+        return $object;
+    }
+
+    /**
+     * @param non-empty-list<string> $paths
+     */
+    public static function createDefault(array $paths): static
```
DanielBadura, line 153: If we would add the registry i would remove this one here and vote for:
```php
$serializer = new JsonSerializer(
    new AttributeEventMetadataFactory(), 
    EventRegistry::fromPaths(['path/to/events'])
);
```

DanielBadura (approved):
Besides the smaller comments this looks good :)

PR #203, opened 2022-03-26
activate renovate automerge

DavidBadura:


DanielBadura:
Why here and not globally?

PR #205, opened 2022-03-28
Transactions

DavidBadura:
With this change you have the possibility to make changes to multiple aggregates transaction safely:

```php
$fromAccount = $this->accountRepository->get($command->fromAccountId);
$toAccount = $this->accountRepository->get($command->toAccountId);

$fromAccount->sendMoney($command->toAccountId, $command->amount);
$toAccount->receiveMoney($command->fromAccountId, $command->amount);

$this->store->transactional(fn () => {
  $this->accountRepository->save($fromAccount);
  $this->accountRepository->save($toAccount);
});
```

### DIFF
--- a/src/Store/DoctrineStore.php
+++ b/src/Store/DoctrineStore.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Patchlevel\EventSourcing\Store;
 
+use Closure;
 use DateTimeImmutable;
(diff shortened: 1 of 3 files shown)


DanielBadura:
I think we should add a testcase for that. I know the concrete implementation is not our side here but i feel like we should add one.

DanielBadura (approved):

PR #209, opened 2022-04-02
rename record method to recordThat

DavidBadura:
Why rename it?

1) It improves the flow of reading
2) record is to general. could be a getter in the business domain.

```php
$this->recordThat(new GuestIsCheckedIn($guestName));
```


DanielBadura (approved):
feels good

PR #223, opened 2022-04-23
hydrator component

### LINKED ISSUE #221: make hydrate and extract private in JsonSerializer
The methods `hydrate` and `extract` are public in JsonSerializer by mistake, they should be set to private before release.

DavidBadura:
move hydrator logic into own class

close #221

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -88,7 +88,7 @@ You also need to implement a `normalize` and `denormalize` method.
 The important thing is that the result of Normalize is serializable.
 
 ```php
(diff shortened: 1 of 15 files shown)


DanielBadura (approved):
Idk but i feel like the name Hydrator is not right here. It really feels more like a normalizer/denormalizer. Besides that 👍

DavidBadura:
We already use the name "normalizer" to normalize individual values. And "hydrator" is a common name for transforming arrays into objects.

In any case, let's look at it again later to see if it can be structured better. I'm rebuilding the whole thing in several stages.

PR #227, opened 2022-04-24
psr container

### LINKED ISSUE #25: add psr container factory
At the moment it's not that easy to explain how to plug the classes together. It would be good for the developer experience if there was a builder or configurator.

Maybe like https://github.com/Roave/psr-container-doctrine

DavidBadura:
- needs #235


DavidBadura:
was outsourced into a separate package: https://github.com/patchlevel/event-sourcing-psr-container

PR #232, opened 2022-04-25
outbox

### LINKED ISSUE #199: outbox
TODO: more description

https://particular.net/webinars/exactly-once-processing-is-easy-isnt-it

DavidBadura:
resolve #199

### DIFF
--- a/src/Console/Command/OutboxConsumeCommand.php
+++ b/src/Console/Command/OutboxConsumeCommand.php
@@ -22,7 +22,7 @@ final class OutboxConsumeCommand extends Command
     {
         $this
             ->setName('event-sourcing:outbox:consume')
-            ->setDescription('show events from one aggregate')
+            ->setDescription('published the messages from the outbox store')
             ->addOption('limit', null, InputOption::VALUE_REQUIRED, 'How many messages should be consumed in one run');
     }
 
--- a/src/Console/Command/OutboxConsumeCommand.php
+++ b/src/Console/Command/OutboxConsumeCommand.php
@@ -4,10 +4,11 @@ declare(strict_types=1);
 
 namespace Patchlevel\EventSourcing\Console\Command;
 
+use Patchlevel\EventSourcing\Console\InputHelper;
 use Patchlevel\EventSourcing\Outbox\OutboxConsumer;
 use Symfony\Component\Console\Command\Command;
-use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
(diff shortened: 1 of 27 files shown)


src/Outbox/StoreOutboxConsumer.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Patchlevel\EventSourcing\Outbox;
+
+use Patchlevel\EventSourcing\EventBus\EventBus;
+use Patchlevel\EventSourcing\Store\OutboxStore;
+
+class StoreOutboxConsumer implements OutboxConsumer
```
DanielBadura, line 10: I think swapping the naming leads to better readability `OutboxConsumerStore`. But i feel like both options are somewhat misleading.

src/Console/Command/OutboxConsumeCommand.php
```diff
@@ -0,0 +1,37 @@
...
+    public function __construct(private OutboxConsumer $consumer)
+    {
+        parent::__construct();
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->setName('event-sourcing:outbox:consume')
+            ->setDescription('show events from one aggregate')
```
DanielBadura: the description here is not right

src/Store/DoctrineStore.php
```diff
@@ -48,6 +56,97 @@ public function connection(): Connection
...
+                        [
+                            'aggregate' => $this->aggregateRootRegistry->aggregateName($message->aggregateClass()),
+                            'aggregate_id' => $message->aggregateId(),
+                            'playhead' => $message->playhead(),
+                        ]
+                    );
+                }
+            }
+        );
+    }
```
DanielBadura, line 132: This is not really "marking" the message as consumed, but more a "delete consumed messages". But i think, this should be fine.

src/Outbox/StoreOutboxConsumer.php
DavidBadura: I think it's a naming convention when you have an interface, like "OutboxConsumer" here, that you just prefix the implementations, so the suffix stays the same. I don't like the name either, but I don't know if your version is better :D

src/Store/DoctrineStore.php
DavidBadura: From the API side you just want to say that the message was processed. But what you then do with the info in the implementation doesn't really matter.

Maybe we should remove the `mark` prefix?

`markOutboxMessageConsumed` => `outboxMessageConsumed`

DanielBadura: Tbh, i dont think so. In our usecase marking is deleting. Maybe someone else implements it different and really only marks the message since they dont want do delete it and loose the data. So i think this is fine as it is.

DanielBadura (approved):

PR #243, opened 2022-05-06
add header for messages

### LINKED ISSUE #234: Message context
We should give the possibility to add context information to the message. A use case is eg. to save which user or system created this message.

The question is also, among other things, how the developer can enrich this information. This has to happen somehow between the repository and the database.

* [x] Aggregate without messages
* [x] Refactor Messages -> payload, header
* [x] Add decorator for messages in repository

DavidBadura:


DanielBadura (approved):
really like the watchserver changes with the new serializer interface 👍

PR #247, opened 2022-05-06
add upcasting possibilty

### LINKED ISSUE #244: Upcasting

DanielBadura:
fix #244

### DIFF
--- a/src/Serializer/DefaultEventSerializer.php
+++ b/src/Serializer/DefaultEventSerializer.php
@@ -44,11 +44,11 @@ final class DefaultEventSerializer implements EventSerializer
     public function deserialize(SerializedEvent $data, array $options = []): object
     {
         $payload = $this->encoder->decode($data->payload, $options);
-        $class = $this->eventRegistry->eventClass($data->name);
+        $upcast = ($this->upcaster)(new Upcast($data->name, $payload));
 
-        $upcast = ($this->upcaster)(new Upcast($class, $payload));
+        $class = $this->eventRegistry->eventClass($upcast->eventName);
 
-        return $this->hydrator->hydrate($upcast->class, $upcast->payload);
+        return $this->hydrator->hydrate($class, $upcast->payload);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/Serializer/DefaultEventSerializer.php
```diff
@@ -38,10 +43,12 @@ public function serialize(object $event, array $options = []): SerializedEvent
 
     public function deserialize(SerializedEvent $data, array $options = []): object
     {
-        $class = $this->eventRegistry->eventClass($data->name);
         $payload = $this->encoder->decode($data->payload, $options);
+        $class = $this->eventRegistry->eventClass($data->name);
+
+        $upcast = ($this->upcaster)(new Upcast($class, $payload));
```
DanielBadura: we could change it to use the `name` instead `class`

src/Serializer/Upcast/Upcast.php
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * @psalm-immutable
+ */
+final class Upcast
+{
+    /**
+     * @param class-string         $class
+     * @param array<string, mixed> $payload
+     */
+    public function __construct(public string $class, public array $payload)
```
DanielBadura: tried to make it as simple as possible, if the move to 8.1 we could make then even readonly

tests/Unit/Serializer/DefaultEventSerializerTest.php
```diff
@@ -49,4 +56,39 @@ public function testDeserialize(): void
...
+        $upcaster = new class implements Upcaster {
+            public function __invoke(Upcast $upcast): Upcast
+            {
+                if ($upcast->class !== ProfileCreated::class) {
+                    return $upcast;
+                }
+
+                return new Upcast($upcast->class, $upcast->payload + ['email' => 'info@patchlevel.de']);
+            }
+        };
```
DanielBadura: idk if this would be a real usecase like i did it here. Also i tried to not add a supports method but then the user must return the original. I'm not sure what is better

src/Serializer/DefaultEventSerializer.php
DavidBadura: I think taking the name brings us more. This also allows you to rename an event without performing a migration. And the mapping still happens afterwards from name to class. So, I would want to convert it to the name.

tests/Unit/Serializer/DefaultEventSerializerTest.php
DavidBadura: I think it's good not to have a support method. 👍

src/Serializer/DefaultEventSerializer.php
DanielBadura: Updated it :)

tests/Unit/Serializer/DefaultEventSerializerTest.php
DanielBadura: Then lets let it be.

PR #248, opened 2022-05-07
bump php versio to 8.1

### LINKED ISSUE #246: bump min php version to 8.1

DavidBadura:


src/EventBus/Message.php
```diff
@@ -8,21 +8,18 @@
 use Patchlevel\EventSourcing\Aggregate\AggregateRoot;
 use Patchlevel\EventSourcing\Clock;
 
-/**
- * @psalm-immutable
```
DanielBadura: i think the immutable marker could stay in. I mean it doesnt harm us, but maybe prevent us from mistakes like adding a mutable property, wdyt?

DanielBadura (approved):

PR #252, opened 2022-05-09
first step of adding decoration possibilities for messages

### LINKED ISSUE #234: Message context

DanielBadura:


src/EventBus/Message.php
```diff
@@ -30,6 +33,12 @@ final class Message
      */
     public function __construct(object $event, array $headers = [])
     {
+        foreach ($headers as $header => $value) {
+            if (!is_string($header)) {
+                throw new RuntimeException('The headers must be an associative array!');
```
DavidBadura: InvalidArgumentException would fit better.

DanielBadura:
LGTM 👍

PR #253, opened 2022-05-09
Add native normalizer

### LINKED ISSUE #218: Add DateTime/DateTimeImmutable normalizer
We should provide normalizers for native data types.

### LINKED ISSUE #219: Add Enum normalizer
We should provide normalizers for native data types.

### LINKED ISSUE #224: Add array normalizer

DavidBadura:
* DateTimeNormalizer
* DateTimeImmutableNormalizer
* DateTimeZoneNormalizer
* EnumNormalizer
* ArrayNormalizer


src/Serializer/Normalizer/DateTimeNormalizer.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        if ($value === null) {
+            return null;
+        }
+
+        if (!is_string($value)) {
+            throw new InvalidArgument();
+        }
+
+        if ($this->immutable) {
```
DanielBadura: im not sure if we should split this into two different ones instead, then we also could typhint the return value better since DateTimeInterface is such a weird thing to typehint. (e.g. no `modify` method is present at the interface)

src/Serializer/Normalizer/ArrayNormalizer.php
```diff
@@ -0,0 +1,41 @@
...
+
+class ArrayNormalizer implements Normalizer
+{
+    public function __construct(private readonly Normalizer $normalizer)
+    {
+    }
+
+    public function normalize(mixed $value): ?array
+    {
+        if ($value === null) {
```
DanielBadura, line 18: Im not sure about the nullability part. Maybe this should be configured so that the normalizer throw if nullability is not allowed? I dont know, if this is somehow better for the user instead of maybe a hard crash, if the property does not allow `null`. Do we have a testcase for this maybe?

DanielBadura (commented):
2 minor comments from my side and more questions to discuss 👍

src/Serializer/Normalizer/ArrayNormalizer.php
DavidBadura: I don't think we should ensure that the value is valid. Whether it's nullable or not, or if you can put a DateTimeNormalizer on a boolean property. We don't do validation here. 

In Doctrine you can also set a column type that doesn't match your property type:

https://github.com/doctrine/dbal/blob/3.3.x/src/Types/DateTimeImmutableType.php

But I will improve the error handling 👍

DanielBadura (approved):
i have one last question which must not be done in here: What about timezone support? Do we need something for this?

DanielBadura (approved):

PR #257, opened 2022-05-19
Clock service and decorator

### LINKED ISSUE #251: clock service

DanielBadura:


src/Repository/DefaultRepository.php
```diff
@@ -129,7 +129,7 @@ static function (object $event) use ($aggregate, &$playhead, $messageDecorator)
                     ->withAggregateClass($aggregate::class)
                     ->withAggregateId($aggregate->aggregateRootId())
                     ->withPlayhead(++$playhead)
-                    ->withRecordedOn(Clock::createDateTimeImmutable());
+                    ->withRecordedOn(new DateTimeImmutable());
```
DanielBadura: i dont know how we should handle this

DavidBadura: i think you can remove this line

DavidBadura: And you need add the RecordedOnDecorator as default decorator.

src/Clock/Clock.php
```diff
@@ -0,0 +1,12 @@
...
+
+declare(strict_types=1);
+
+namespace Patchlevel\EventSourcing\Clock;
+
+use DateTimeImmutable;
+
+interface Clock
+{
+    public function createDateTimeImmutable(): DateTimeImmutable;
```
DanielBadura: Maybe we should also rename this to only `create`, wdyt?

src/Clock/FreezeClock.php
```diff
@@ -0,0 +1,34 @@
...
+
+use function sprintf;
+
+final class FreezeClock implements Clock
+{
+    public function __construct(private DateTimeImmutable $frozenDateTime)
+    {
+    }
+
+    public function freeze(DateTimeImmutable $frozenDateTime): void
```
DanielBadura: im unsure about this, but i think in testing this could be really helpful to freeze with another date later on.

src/Repository/DefaultRepository.php
DanielBadura: Yeah but where? Should we add this at the manager if no decorator is given? What happen then if a decorator is given but no `RecordedOnDecorator` is in the chain e.g.

src/Clock/FreezeClock.php
DavidBadura: Yes, that is definitely useful. Maybe it would be better to name the method: `changeFreezedDateTime`.

src/Repository/DefaultRepository.php
DavidBadura: I would set it as default in the repository. If `RecordedOnDecorator` is not set, then there is an error when writing that the RecordedOn header could not be found.

All systems with default middleware or decorators have the problem that the user may overwrite them or forget to include them. eg. with the message component you can also forget about the HandleMiddleware. And then the whole component no longer makes sense.

DavidBadura (approved):

PR #259, opened 2022-06-07
Silence socket warnings for DefaultWatchServerClient

robinlehrmann:
Event if I deactivate the watch server in the symfony config it executes this watcher and I get some warning, that the socket is not available or broken pipe warnings. What do you think to prevent warnings like this?

### DIFF
--- a/src/WatchServer/DefaultWatchServerClient.php
+++ b/src/WatchServer/DefaultWatchServerClient.php
@@ -53,7 +53,7 @@ final class DefaultWatchServerClient implements WatchServerClient
         set_error_handler([self::class, 'nullErrorHandler']);
 
         try {
-            if (stream_socket_sendto($socket, $encodedPayload) !== -1) {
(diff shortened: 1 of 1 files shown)


DanielBadura:
I'm not in favor of silencing the warnings. But could you share the warnings and your config?

DavidBadura:
The watcher was inspired from Symfony. Symfony silence this warnings too. If i create the watcher i think about it and for the first, i remove the silence operator. But after rethinking, it make sense to silence it. It is annoying to activate and deactivate the listener over the config, if you don't use the watch server currently. So for me it is 👍 to silence it.

But we have a bug in the DI. The watch listener should be registered if you deactivate the watch server in the configuration. I open a bug ticket for it: https://github.com/patchlevel/event-sourcing-bundle/issues/89

DavidBadura (approved):

DanielBadura:
Thanks @robinlehrmann ! :)

PR #274, opened 2022-06-24
remove fqcn support in normalize attribute

DavidBadura:


DanielBadura (changes requested):
You need to also update `normalizer.md`

DanielBadura (approved):

PR #277, opened 2022-06-27
Fix README.md

robinlehrmann:


README.md
```diff
@@ -12,11 +12,11 @@ A lightweight but also all-inclusive event sourcing library with a focus on deve
 * Everything is included in the package for event sourcing
 * Based on [doctrine dbal](https://github.com/doctrine/dbal) and their ecosystem
 * Developer experience oriented and fully typed
-* [Snapshots](docs/snapshots.md) system to quickly rebuild the aggregates
```
DavidBadura, line 15: good catch!

But we now have our own documentation website to which we should now link.

For example for snapshots: 

https://patchlevel.github.io/event-sourcing-docs/2.0/snapshots/

The documentation is not fully updated yet, we are working on it. This is the last step before the 2.0 release.

DanielBadura:
This got fixed with #279, thanks for reporting it @robinlehrmann! Closing this here therefore.

PR #279, opened 2022-06-28
finish docs for 2.0

DavidBadura:
needs #280

### DIFF
--- a/docs/pages/aggregate.md
+++ b/docs/pages/aggregate.md
@@ -589,7 +589,7 @@ final class Profile extends AggregateRoot
 }
 ```
 
-But if you still want to make sure that the time is now and not in the past or future, you can pass a clock.
+But if you still want to make sure that the time is "now" and not in the past or future, you can pass a clock.
 
 ```php
 use Patchlevel\EventSourcing\Aggregate\AggregateRoot;
@@ -616,7 +616,8 @@ final class Profile extends AggregateRoot
 }
 ```
 
-This gives you the option of implementing the clock, e.g. use the FrozenClock to always get the same time.
+Now you can pass the `SystemClock` to determine the current time. 
+Or for test purposes the `FrozenClock`, which always returns the same time.
 
 !!! note
 
--- a/docs/pages/aggregate.md
+++ b/docs/pages/aggregate.md
@@ -7,7 +7,7 @@
 
     [DDD Aggregate - Martin Flower](https://martinfowler.com/bliki/DDD_Aggregate.html) 
 
-An AggregateRoot has to inherit from `AggregateRoot` and need to implement the method `aggregateRootId`.
+An Aggregate has to inherit from `AggregateRoot` and need to implement the method `aggregateRootId`.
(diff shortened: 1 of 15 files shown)


docs/pages/aggregate.md
```diff
@@ -559,6 +557,71 @@ final class Hotel extends AggregateRoot
...
+        $self->recordThat(new ProfileRegistered($id, $name, $clock->now()));
+
+        return $self;
+    }
+    
+    // ...
+}
+```
+
+This gives you the option of implementing the clock, e.g. use the FrozenClock to always get the same time.
```
DanielBadura: Maybe add a note for testing here in this context. To prevent ppl. from thinking this should be done in production with the FrozenClock :D

docs/pages/pipeline.md
```diff
@@ -303,3 +326,55 @@ $middleware = new ChainMiddleware([
     new RecalculatePlayheadMiddleware()
 ]);
 ```
+
+### Custom middleware
+
+You can also write a custom middleware. The middleware gets a message and can return n messages.
```
DanielBadura:
```suggestion
You can also write a custom middleware. The middleware gets a message and can return `N` messages. 
```

docs/pages/repository.md
```diff
@@ -35,38 +32,76 @@ $repository = $repositoryManager->get(Profile::class);
...
+    The same repository instance is always returned for a specific aggregate.
+
+### Snapshots
 
-### Default Repository Manager
+Loading events for an aggregate is superfast. 
+You can have thousands of events in the database that load in a few milliseconds and build the corresponding aggregate.
 
-// todo
+But at some point you realize that it takes time. To counteract this there is a snapshot store.
```
DanielBadura, line 42: Would add a note about this is behaving like a cache here maybe, or add a warning to make the user aware, even if he is not reading the snapshot doc? Im not sure if needed, just wanted to question it.

docs/pages/snapshots.md
```diff
@@ -15,65 +15,110 @@ Here, however, only the last events are loaded from the database and not all.
...
-## Batch
+!!! note
+
+    You can find more about normalizer [here](normalizer.md).
+
+### Snapshot batching
+
+Since the loading of events in itself is quite fast and only becomes noticeably slower with thousands of events, 
+we do not need to create a snapshot after each event. That would also have a negative impact on performance. 
+Instead, we can also create a snapshot after `n` events.
```
DanielBadura:
```suggestion
Instead, we can also create a snapshot after `N` events. 
```

DanielBadura (approved):

docs/pages/repository.md
DavidBadura: Just adding the SnapshotStore in DefaultRepositoryManager is not enough to make it cached. You need to read the snapshot store documentation to know how to setup the snapshot store. 

If you have a short sentence without describing much around it, you're welcome. But otherwise i am against much more to describe at this point.

DanielBadura (approved):

PR #291, opened 2022-07-02
snapshots version

### LINKED ISSUE #226: Snapshot Version

DavidBadura:
allow snapshots to be versioned using an attribute. If the version doesn't match what is loaded from the store, then the snapshot is discarded and the aggregate is built from scratch.

fix #226

### DIFF
--- a/src/Metadata/AggregateRoot/AggregateRootMetadata.php
+++ b/src/Metadata/AggregateRoot/AggregateRootMetadata.php
@@ -17,7 +17,7 @@ final class AggregateRootMetadata
         public readonly bool $suppressAll,
         public readonly ?string $snapshotStore,
         public readonly ?int $snapshotBatch,
-        public readonly int|string|null $snapshotVersion = null,
+        public readonly ?string $snapshotVersion = null,
     ) {
     }
 }
--- a/src/Metadata/AggregateRoot/AggregateRootMetadata.php
+++ b/src/Metadata/AggregateRoot/AggregateRootMetadata.php
@@ -17,6 +17,7 @@ final class AggregateRootMetadata
         public readonly bool $suppressAll,
         public readonly ?string $snapshotStore,
         public readonly ?int $snapshotBatch,
+        public readonly int|string|null $snapshotVersion = null,
     ) {
     }
 }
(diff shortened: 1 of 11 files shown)


src/Snapshot/SnapshotExpired.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Patchlevel\EventSourcing\Snapshot;
+
+use function sprintf;
+
+final class SnapshotExpired extends SnapshotException
```
DavidBadura: @DanielBadura maybe you have a better name for it. I think "expired" doesn't quite fit.

DanielBadura: Maybe call it for what it really stand for: SnapshotVersionInvalid

src/Metadata/AggregateRoot/AggregateRootMetadata.php
```diff
@@ -17,6 +17,7 @@ public function __construct(
         public readonly bool $suppressAll,
         public readonly ?string $snapshotStore,
         public readonly ?int $snapshotBatch,
+        public readonly int|string|null $snapshotVersion = null,
```
DanielBadura: Why do we also allow string? Do we expect such large numbers which won't fit in int?

src/Metadata/AggregateRoot/AttributeAggregateRootMetadataFactory.php
```diff
@@ -50,8 +50,9 @@ public function metadata(string $aggregate): AggregateRootMetadata
             $properties,
             $suppressEvents,
             $suppressAll,
-            $snapshotStore,
-            $snapshotBatch,
+            $snapshot?->name(),
+            $snapshot?->batch(),
+            $snapshot?->version()
```
DanielBadura, line 55: maybe just pass the array for the snapshot and define the shape `array{name: string, batch: int, version: string|int|null}`?

src/Metadata/AggregateRoot/AggregateRootMetadata.php
DavidBadura: I would always use uuids, that's much easier to manage. You don't have to be careful that you count up correctly. But I still want to give the possibility to use a number.

DanielBadura: So maybe just drop `int`? You could pass it also like `'1'` if someone wants a number or open it up later on if requested.

src/Metadata/AggregateRoot/AttributeAggregateRootMetadataFactory.php
DavidBadura: I cant'? this is a BC break

DanielBadura: True, thought it was internal but it isnt.

DanielBadura (approved):

PR #293, opened 2022-07-08
add worker for cli commands

### LINKED ISSUE #292: worker for outbox consumer
Currently we don't have a cli worker implementation. But we need that in some places. To one for our outbox consumer. But also in the future for projections.

The worker should run all the time. But you can set when it should shut down.

- Number of processed items
- Memory limit exceeded
- Time exceeded

The interception of ctrl+c should also be implemented so that it is only terminated after the item and not immediately.

DavidBadura:
* resolve #292

This PR adds a worker system, which is also used by the outbox consumer.

You can configure the worker and define one or more stop rules:

* When the memory limit has been reached
* After a certain time
* After a certain number of runs
* When SIGTERM is sent (ctrl+c)

The process is ended in the middle, but only at the end of a run.

Furthermore, you can also set how many ms should be waited between the runs.

### DIFF
--- a/baseline.xml
+++ b/baseline.xml
@@ -1,9 +1,24 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <files psalm-version="4.24.0@06dd975cb55d36af80f242561738f16c5f58264f">
+  <file src="src/Console/Command/OutboxConsumeCommand.php">
+    <ArgumentTypeCoercion occurrences="1">
+      <code>$sleep</code>
+    </ArgumentTypeCoercion>
+  </file>
   <file src="src/Console/DoctrineHelper.php">
     <ClassNotFinal occurrences="1">
       <code>class DoctrineHelper</code>
     </ClassNotFinal>
+  </file>
+  <file src="src/Console/Worker/DefaultWorker.php">
+    <TypeDoesNotContainType occurrences="1">
+      <code>$this-&gt;shouldStop</code>
+    </TypeDoesNotContainType>
+  </file>
+  <file src="src/Console/Worker/Listener/StopWorkerOnSigtermSignalListener.php">
+    <MoreSpecificReturnType occurrences="1">
+      <code>array&lt;class-string, string&gt;</code>
+    </MoreSpecificReturnType>
   </file>
   <file src="src/Repository/DefaultRepository.php">
     <MixedArgument occurrences="1">
--- a/src/Console/Command/OutboxConsumeCommand.php
+++ b/src/Console/Command/OutboxConsumeCommand.php
@@ -5,12 +5,21 @@ declare(strict_types=1);
 namespace Patchlevel\EventSourcing\Console\Command;
 
 use Patchlevel\EventSourcing\Console\InputHelper;
+use Patchlevel\EventSourcing\Console\InvalidArgumentGiven;
+use Patchlevel\EventSourcing\Console\Worker\Bytes;
+use Patchlevel\EventSourcing\Console\Worker\DefaultWorker;
+use Patchlevel\EventSourcing\Console\Worker\Listener\StopWorkerOnIterationLimitListener;
+use Patchlevel\EventSourcing\Console\Worker\Listener\StopWorkerOnMemoryLimitListener;
+use Patchlevel\EventSourcing\Console\Worker\Listener\StopWorkerOnSigtermSignalListener;
+use Patchlevel\EventSourcing\Console\Worker\Listener\StopWorkerOnTimeLimitListener;
 use Patchlevel\EventSourcing\Outbox\OutboxConsumer;
 use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Logger\ConsoleLogger;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\EventDispatcher\EventDispatcher;
 
 #[AsCommand(
     'event-sourcing:outbox:consume',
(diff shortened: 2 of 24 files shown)


src/Console/Worker/Listener/StopWorkerOnTimeLimitListener.php
```diff
@@ -0,0 +1,53 @@
...
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+use function microtime;
+
+final class StopWorkerOnTimeLimitListener implements EventSubscriberInterface
+{
+    private float $endTime = 0;
+
+    public function __construct(
+        private readonly int $timeLimitInSeconds,
```
DavidBadura, line 22:
```suggestion
        private readonly int $timeLimitInMicroSeconds,
```

tests/Unit/Console/Worker/BytesTest.php
```diff
@@ -0,0 +1,44 @@
...
+     * @dataProvider validParseDataProvider
+     */
+    public function testValidParse(string $string, int $expectedBytes): void
+    {
+        $bytes = Bytes::parseFromString($string);
+
+        self::assertSame($expectedBytes, $bytes->value());
+    }
+
+    public function validParseDataProvider(): Generator
```
DanielBadura, line 41: Add the type here so that we dont need to add the error to the baseline?

src/Console/Worker/DefaultWorker.php
```diff
@@ -0,0 +1,77 @@
...
+
+            ($this->job)();
+
+            $ranTime = (int)(microtime(true) - $startTime);
+
+            $this->logger?->debug('Worker finished job run ({ranTime}ms)', ['ranTime' => $ranTime]);
+
+            $this->eventDispatcher->dispatch(new WorkerRunningEvent($this));
+
+            if ($this->shouldStop) {
```
DanielBadura, line 48: is this micro optimization needed? since we are just skipping the sleep duration if given, right?

baseline.xml
```diff
@@ -5,6 +5,16 @@
       <code>class DoctrineHelper</code>
     </ClassNotFinal>
   </file>
+  <file src="src/Console/Worker/DefaultWorker.php">
+    <TypeDoesNotContainType occurrences="1">
+      <code>$this-&gt;shouldStop</code>
```
DanielBadura, line 15: i dont understand this one. What exactly is here wrong?

src/Console/Command/OutboxConsumeCommand.php
```diff
@@ -5,35 +5,122 @@
...
+
+        if ($runLimit) {
+            if ($runLimit <= 0) {
+                throw new InvalidArgumentGiven($runLimit, 'null|positive-int');
+            }
+
+            $eventDispatcher->addSubscriber(new StopWorkerOnIterationLimitListener($runLimit, $logger));
+        }
+
+        if ($memoryLimit) {
```
DanielBadura, line 107: should we here also check for positive-int? And if yes, maybe it makes sense to move it to the InputHelper.

src/Console/Command/OutboxConsumeCommand.php
```diff
@@ -5,35 +5,122 @@
...
+            $eventDispatcher->addSubscriber(new StopWorkerOnMemoryLimitListener(Bytes::parseFromString($memoryLimit), $logger));
+        }
+
+        if ($timeLimit) {
```
DanielBadura, line 111: same here i think

src/Console/Worker/Bytes.php
```diff
@@ -0,0 +1,46 @@
...
+
+final class Bytes
+{
+    /** @var array<string, int> */
+    private static array $sizes = [
+        'B' => 1,
+        'KB' => 1_024,
+        'MB' => 1_048_576,
+        'GB' => 1_073_741_824,
+    ];
```
DanielBadura: Maybe use a const instead?
```php
    private const SIZES = [
        'B' => 1,
        'KB' => 1_024,
        'MB' => 1_048_576,
        'GB' => 1_073_741_824,
    ];
```

src/Console/Worker/Listener/StopWorkerOnSigtermSignalListener.php
```diff
@@ -0,0 +1,42 @@
...
+final class StopWorkerOnSigtermSignalListener implements EventSubscriberInterface
+{
+    public function __construct(
+        private readonly ?LoggerInterface $logger = null
+    ) {
+    }
+
+    public function onWorkerStarted(WorkerStartedEvent $event): void
+    {
+        pcntl_signal(SIGTERM, function () use ($event): void {
```
DanielBadura, line 25: Shouldnt we here also check for the function existing?

src/Console/Worker/Listener/StopWorkerOnTimeLimitListener.php
DanielBadura, line 22: maybe also positive-int?

src/Console/Worker/Worker.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Patchlevel\EventSourcing\Console\Worker;
+
+interface Worker
+{
+    /**
+     * @param int $sleepTimer sleepTimer in microseconds
```
DanielBadura: here we should also consider `0|positive-int`

DanielBadura (commented):
Some nitpicking about types but besides that it looks really good!

tests/Unit/Console/Worker/BytesTest.php
DavidBadura: if you know how, go for it. try it :D

DanielBadura: Cant reproduce it in psalm.dev but this should be enough i think:
```php
    /**
     * @return Generator<array{0:string, 1:int}>
     */
```

src/Console/Command/OutboxConsumeCommand.php
DavidBadura: This is a string and will be parsed to an integer in `Bytes::parseFromString`. And only positive numbers can be parsed.

DanielBadura (approved):

PR #297, opened 2022-07-22
add projector and projectionist

### LINKED ISSUE #295: new projection system
## new projection system requirement

* Projections should be able to be running asynchronously using a worker (#293)
* The position of a projection should be stored in a store so that the system always knows where a projection is and can continue from there
* It should be possible to prepare new projections when deploying, so that you don't have to wait long for the current projection when deploying (switching the application).
* New versions of a projection should be able to be built in parallel without changing the existing projection and affecting the running application.
* Old projections that are no longer needed (stale) should be able to be cleaned up.
* Faulty projections should be marked. This projection is skipped so that other projections can continue anyway.
* Faulty projections can either be restarted directly (from the current position) or be rebuilt from scratch.
* In general, you should be able to rebuild every projection from scratch.

## inspired blog post

* https://barryosull.com/blog/projection-building-blocks-what-you-ll-need-to-build-projections/
* https://barryosull.com/blog/managing-projectors-is-harder-than-you-think/

DavidBadura:
This is a proof of concept for ticket #295


DavidBadura:
it is on hold and waiting for #317

baseline.xml
```diff
@@ -43,6 +66,86 @@
...
+      <code>Projection</code>
+      <code>Projection</code>
+      <code>Projection</code>
+    </DeprecatedClass>
+  </file>
+  <file src="src/Projection/ProjectorStore/ProjectorStateCollection.php">
+    <ImpureFunctionCall occurrences="1">
+      <code>array_filter</code>
+    </ImpureFunctionCall>
+  </file>
```
DanielBadura: Impure function call? Is the callable mutating the state?

src/Projection/DefaultProjectionist.php
```diff
@@ -0,0 +1,226 @@
...
+        ProjectorCriteria $criteria = new ProjectorCriteria(),
+        ?LoggerInterface $logger = null
+    ): void {
+        $projectorStates = $this->projectorStates()->filterByProjectorStatus(ProjectorStatus::Outdated);
+
+        foreach ($projectorStates as $projectorState) {
+            $projector = $this->projectorRepository->findByProjectorId($projectorState->id());
+
+            if (!$projector) {
+                $logger?->warning(sprintf('projector witt the id "%s" not found', $projectorState->id()->toString()));
```
DanielBadura:
```suggestion
                $logger?->warning(sprintf('projector with the id "%s" not found', $projectorState->id()->toString()));
```

src/Projection/ProjectorStore/DatabaseStore.php
```diff
@@ -0,0 +1,130 @@
...
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\DBAL\Types\Types;
+use Patchlevel\EventSourcing\Projection\ProjectorId;
+use Patchlevel\EventSourcing\Projection\ProjectorStatus;
+use Patchlevel\EventSourcing\Schema\SchemaConfigurator;
+
+use function array_map;
+
+final class DatabaseStore implements ProjectorStore, SchemaConfigurator
```
DanielBadura: Maybe change the naming to `DBALStore` or `DoctrineStore`? Since it's more that then a general `DatabaseStore`.

DanielBadura (approved):

PR #299, opened 2022-07-29
Fix problem with database names using hyphens

icanhazstring:
Hi. First off all. Awesome lib. Thanks <3

While working to get it running, I noticed that there is a problem when creating a database where the name has hyphens in it. Using this, the name must be used with backticks.

### DIFF
--- a/src/Console/DoctrineHelper.php
+++ b/src/Console/DoctrineHelper.php
@@ -60,11 +60,15 @@ class DoctrineHelper
 
     public function createDatabase(Connection $connection, string $databaseName): void
     {
-        $connection->createSchemaManager()->createDatabase($databaseName);
+        $connection->createSchemaManager()->createDatabase(
+            $connection->getDatabasePlatform()->quoteSingleIdentifier($databaseName)
+        );
     }
 
     public function dropDatabase(Connection $connection, string $databaseName): void
     {
-        $connection->createSchemaManager()->dropDatabase($databaseName);
+        $connection->createSchemaManager()->dropDatabase(
+            $connection->getDatabasePlatform()->quoteSingleIdentifier($databaseName)
+        );
     }
 }
--- a/tests/Unit/Console/DoctrineHelperTest.php
+++ b/tests/Unit/Console/DoctrineHelperTest.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Patchlevel\EventSourcing\Tests\Unit\Console;
 
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Platforms\MySQLPlatform;
 use Doctrine\DBAL\Schema\AbstractSchemaManager;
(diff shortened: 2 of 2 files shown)


icanhazstring:
Will squash/rebase if build is green to have a single commit :)

src/Console/DoctrineHelper.php
```diff
@@ -60,11 +61,11 @@ public function hasDatabase(Connection $connection, string $databaseName): bool
 
     public function createDatabase(Connection $connection, string $databaseName): void
     {
-        $connection->createSchemaManager()->createDatabase($databaseName);
+        $connection->createSchemaManager()->createDatabase(sprintf('`%s`', $databaseName));
```
DavidBadura:
```suggestion
        $connection->createSchemaManager()->createDatabase(
            $connection->getDatabasePlatform()->quoteSingleIdentifier($databaseName)
        );
```

src/Console/DoctrineHelper.php
```diff
@@ -60,11 +61,11 @@ public function hasDatabase(Connection $connection, string $databaseName): bool
...
     }
 
     public function dropDatabase(Connection $connection, string $databaseName): void
     {
-        $connection->createSchemaManager()->dropDatabase($databaseName);
+        $connection->createSchemaManager()->dropDatabase(sprintf('`%s`', $databaseName));
```
DavidBadura:
```suggestion
        $connection->createSchemaManager()->dropDatabase(
            $connection->getDatabasePlatform()->quoteSingleIdentifier($databaseName)
        );
```

DavidBadura (commented):
Hey, first of all thanks and good catch! We seem to have missed this case.

I looked again at how it was solved in the `CreateDatabaseCommand` of doctrine bundle and found these lines:

https://github.com/doctrine/DoctrineBundle/blob/2.7.x/Command/CreateDatabaseDoctrineCommand.php#L114

I would also do the quoting via dbal, since i don't know which databases need special treatment :D

In the `CreateDatabaseCommand` it is checked whether a path exists, if so, the quoting is skipped. I don't know if it is relevant to us. I think it only happens with sqlite?

icanhazstring:
Good point absolutely. Was thinking about it also.

> In the CreateDatabaseCommand it is checked whether a path exists, if so, the quoting is skipped. I don't know if it is relevant to us. I think it only happens with sqlite?

Yes. I think this is only sqlite where the db is stored as a file.

DavidBadura:
Okay, i would ignore the "path" case for now. If the pipeline is green i would merge it. 👍 
Oh, can you change the target branch to 2.0.x since it is a bug? Would then ship it with version 2.0.2

icanhazstring:
Squashed and rebased onto 2.0.x 👍

DavidBadura (approved):

DavidBadura:
Thank you!

PR #301, opened 2022-07-29
Add interactive capability for show command

### LINKED ISSUE #233: interactive cli show command
For DX we can make the show command more interactive. If no argument is passed, the aggregate is asked it for first. Here we can display a list of aggregates from which he can select one. Then you will be asked for the aggregate ID. 
This should only happen as an option if no arguments is passed, otherwise it will use the arguments.

icanhazstring:
This should resolve #233

### DIFF
--- a/src/Console/Command/ShowCommand.php
+++ b/src/Console/Command/ShowCommand.php
@@ -14,7 +14,10 @@ use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ChoiceQuestion;
+use Symfony\Component\Console\Question\Question;
 
+use function array_values;
 use function count;
 use function sprintf;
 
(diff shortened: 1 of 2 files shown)


src/Console/Command/ShowCommand.php
```diff
@@ -40,16 +43,30 @@ public function __construct(Store $store, EventSerializer $serializer, Aggregate
...
     }
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $console = new OutputStyle($input, $output);
 
-        $aggregate = InputHelper::string($input->getArgument('aggregate'));
-        $id = InputHelper::string($input->getArgument('id'));
+        $aggregate = InputHelper::nullableString($input->getArgument('aggregate'));
+        while ($aggregate === null) {
```
DanielBadura: Isnt an `if` sufficient for both arguments?

icanhazstring: If you hit just enter you will be asked again until you cancel the command, or enter something.

DanielBadura: I'm not sure if this approach or a error message which hints the user that he needs to choose one option to proceed is better DX.

But lets keep it as is with the requestioning until a value is choosen, maybe with an additional text hinting the user that no input was given? This is up to you if you want to put more effort into it. 

A Testcase covering that the arguments are now optional would be great.

icanhazstring: Good point, will add a hint and test 😉

icanhazstring:
I removed the `while` statement as its not necessary. The question helper will do its part here.
Also added tests using `CommandTester`. If this is not wanted I can refactor it to be structured as the rest of the tests :)

DanielBadura (approved):
This looks good, thank you @icanhazstring!

PR #303, opened 2022-07-30
Add event attribute to split the stream

### LINKED ISSUE #287: Add Events which enables to split the eventstream to keep it short
Right now we have no possibility to split up the stream for an aggregate. This could be handy if the stream gets really long, or even if the business needs some kind of splitting it up. For the first concern we have the snapshots in place, but this should be seen as an cache layer, so no real "shortening" of the stream is done and needs to be invalidated as soon as something is changed for the aggregate internals, there comes https://github.com/patchlevel/event-sourcing/issues/226 into play.

So we could introduce a new Attribute which signals that a split should be done before the event is proccessed. Then we could split on specific events which like could occur monthly or so (tbd by business) which will trigger this. The Store must handle then the multiple streams or we could define that these splitting events need all data an aggregate needs to be re-created, kina like a snapshot. Then we could archive the stream before and move it somewhere else or just mark it as archived via Metadata. This would result then in a big archived stream which is not used for re-creating the aggregate and a small not archived which will be used.

DanielBadura:
Add the possibility to split the stream on a given event with the attribute SplitStream. If an event with this marker attribute is going to be saved, then all past events get archived. Then only the non archived events get loaded to rebuild the aggregate. That means that those marked events need to hold all data needed to re-create the aggregate to the same state as it is in the moment. If this is not given the whole system is going to break.

Why are we doing this? There is for one a performance boost if we are having really big stream for aggregates it could take some time to recreate them if we are going through millions of events. With this we can like start again from one event. Some businesses also require some kind of monthly or yearly reset of the stream.

### DIFF
--- a/src/Attribute/SplitStream.php
+++ b/src/Attribute/SplitStream.php
@@ -1,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Patchlevel\EventSourcing\Attribute;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_CLASS)]
+final class SplitStream
+{
+}
--- a/src/Store/SingleTableStore.php
+++ b/src/Store/SingleTableStore.php
@@ -17,6 +17,7 @@ use Patchlevel\EventSourcing\Serializer\SerializedEvent;
 use function array_map;
 use function is_int;
 use function is_string;
+use function sprintf;
 
 final class SingleTableStore extends DoctrineStore implements PipelineStore
 {
@@ -45,7 +46,10 @@ final class SingleTableStore extends DoctrineStore implements PipelineStore
         $sql = $this->connection->createQueryBuilder()
             ->select('*')
             ->from($this->storeTableName)
-            ->where('aggregate = :aggregate AND aggregate_id = :id AND playhead > :playhead')
+            ->where('aggregate = :aggregate')
+            ->andWhere('aggregate_id = :id')
+            ->andWhere('playhead > :playhead')
+            ->andWhere('archived = :archived')
             ->getSQL();
 
         /** @var list<array{aggregate_id: string, playhead: string|int, event: string, payload: string, recorded_on: string, custom_headers: string}> $result */
@@ -55,6 +59,10 @@ final class SingleTableStore extends DoctrineStore implements PipelineStore
                 'aggregate' => $shortName,
                 'id' => $id,
                 'playhead' => $fromPlayhead,
+                'archived' => false,
+            ],
+            [
+                'archived' => Types::BOOLEAN,
             ]
         );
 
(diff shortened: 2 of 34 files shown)


src/Attribute/SplitStream.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Patchlevel\EventSourcing\Attribute;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_CLASS)]
+final class SplitStream
```
DanielBadura, line 10: Not sure where to save this information

src/Store/SingleTableStore.php
```diff
@@ -110,8 +156,12 @@ public function save(Message ...$messages): void
             function (Connection $connection) use ($messages): void {
                 foreach ($messages as $message) {
                     $event = $message->event();
-
                     $data = $this->serializer->serialize($event);
+                    $isSplitEvent = true;
```
DanielBadura: Here, the information is needed if the event is a splitting one.

src/Store/SingleTableStore.php
```diff
@@ -220,4 +271,39 @@ public function schema(): Schema
...
+            $this->storeTableName,
+            [
+                'archived' => $archivedMessage->archived(),
+            ],
+            [
+                'aggregate' => $this->aggregateRootRegistry->aggregateName($archivedMessage->aggregateClass()),
+                'aggregate_id' => $archivedMessage->aggregateId(),
+                'playhead' => $archivedMessage->playhead(),
+            ]
+        );
```
DanielBadura: With this, we should be really careful. Since we dont want to update the stream. So maybe another solution should be taken? Like creating a new stream and let the old one in the db, like versioning.

DanielBadura (commented):
By far not finished, but wanted to share it already.

src/Store/SingleTableStore.php
```diff
@@ -220,4 +271,39 @@ public function schema(): Schema
...
+
+        foreach ($toBeArchivedMessages as $toBeArchivedMessage) {
+            if ($toBeArchivedMessage->archived()) {
+                // already archived, should we exit here since there seems something wrong with the
+                // query?
+            }
+
+            $this->archiveMessage($toBeArchivedMessage);
+        }
+    }
```
DavidBadura: why don't you create an update statement that simply sets the flag "archive" for all messages of an aggregate up to this playhead?

```sql
UPDATE table_name
SET archived = 1
WHERE
    aggregate_id = :aggregateId
    playhead < :playhead
    archived = 0
```

So i would not want to reload the whole stream to edit the message and then write it again.

src/Attribute/SplitStream.php
DavidBadura, line 10: I don't like the name of this attribute. How about?

* RebootStream
* RestartStream

DavidBadura: Or "NewStart" ?

DanielBadura: Well, i think all the proposed names are not good either tbh. 

* RebootStream feels off here
* RestartStream feels more like reloading
* NewStart meh - i dont know if the user can understand this

We should first think about how we want the feature to be in the future. Do we want to stay with the `achieved` bool flag or do we want to *really* split / restart the stream? 

For me SplitStream is the most accurate, as we are already using the stream term in some places in docs and code. I think this is the most understandable name from the 4 options right now. But we can put more thought about the name for sure.

DanielBadura: Renamed it on the message to be more clear in the message context what the flag is supposed to do. Will resolve this for now with that.

DavidBadura (approved):

PR #305, opened 2022-08-03
Fix current version normalization

### LINKED ISSUE #302: Documentation update workflow is currently not working correctly
See: https://github.com/patchlevel/event-sourcing/runs/7579745095?check_suite_focus=true#step:7:28

icanhazstring:
This will allow the `docs-build` workflow to work with branch names like `0.0.x` and also tagged releases like `0.0.0`.
Should solve #302

### DIFF
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -41,7 +41,7 @@ jobs:
         id: current
         uses: actions/github-script@v6
         with:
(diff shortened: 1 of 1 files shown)


DanielBadura (approved):
Regex is looking good for me. Testing this would be hard, i think it will be tested with the next release :D

Will also port this fix to the bundle.

DanielBadura:
Thanks @icanhazstring 👍

PR #324, opened 2022-10-25
Solve issue with adding events to bus

icanhazstring:
This will fix an issue that happens if during an event listener another event might end up in the queue. This is because `array += array` does behave in another way than what might be expected.

Simple reproduction using php: https://3v4l.org/bGcG3#v8.1.11 Here `array_merge` would be the correct way.

### DIFF
--- a/src/EventBus/DefaultEventBus.php
+++ b/src/EventBus/DefaultEventBus.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Patchlevel\EventSourcing\EventBus;
 
+use function array_merge;
 use function array_shift;
 
 final class DefaultEventBus implements EventBus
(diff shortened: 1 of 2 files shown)


icanhazstring:
Will try and add a valid test to reproduce the problem.

DanielBadura:
Change itself looks good to me. You can ignore the failing BC-Check Job.

icanhazstring:
Added a test. The problem occurs if we are in the middle of processing multiple messages. If one`$listener` adds another message, this doesn't get added to the list. Using `array_merge` the new messages are added at the end of `$this->queue` and get fetched using `array_shift`.

After that, all `$listener` are executed.

icanhazstring:
Should I target 2.0.x so that we can do upward merge into 2.1.x?

DavidBadura:
Thanks for the fix! And yes, please change the target to 2.0.x

icanhazstring:
> please change the target to 2.0.x

Done :)

DavidBadura:
thank you!

PR #332, opened 2022-11-25
add projector helper

DavidBadura:


src/Projection/ProjectorHelper.php
```diff
@@ -0,0 +1,69 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Patchlevel\EventSourcing\Projection;
+
+use Patchlevel\EventSourcing\EventBus\Message;
+
+final class ProjectorHelper
```
DanielBadura, line 9: The naming is not hinting the user what the class is really doing. Can we get a better name here?

src/Projection/ProjectorHelper.php
```diff
@@ -0,0 +1,69 @@
...
+
+    public function createProjectionUsingRepository(ProjectorRepository $projectorRepository): void
+    {
+        $this->createProjection(...$projectorRepository->projectors());
+    }
+
+    public function dropProjectionUsingRepository(ProjectorRepository $projectorRepository): void
+    {
+        $this->dropProjection(...$projectorRepository->projectors());
+    }
```
DanielBadura: Are these really needed? Imho they are not adding much value on top the other methods.

DanielBadura (changes requested):
I like how the new `ProjectionHelper` is now a real replacement to the old `MetadataAwareProjectionHandler`. But as i mentioned above the naming should be improved.

src/Projection/ProjectorHelper.php
DavidBadura: Do you have a suggestion?

src/Projection/ProjectorHelper.php
DavidBadura: Makes it easier to use. And since it's not an interface and just a utility class, I think such small improvements can be added.

src/Projection/ProjectorHelper.php
DanielBadura: Hm idk, creating a issue for that to think about it again before releasing `2.1.0`

DanielBadura (approved):

PR #333, opened 2022-11-25
remove from option so that the library handles the base branch

DanielBadura:
As long as we use semver branches and tags this should work. See https://github.com/Roave/BackwardCompatibilityCheck#usage


DavidBadura:
needs rebase :)

DavidBadura (approved):

PR #334, opened 2022-11-25
support PSR-20 ClockInterface

DavidBadura:
https://github.com/php-fig/clock


DanielBadura (approved):

DanielBadura (changes requested):
About the BC-Break, will try to investigate this and maybe report upstream.

DanielBadura (approved):
As I mentioned reproduced the wrong BC-Break upstream and workin on a fix for that. So we can merge this here since it's not a bc-break.

PR #337, opened 2022-12-05
refactor projectionist and add documentation

### LINKED ISSUE #320: projector-id filter option for all projectionist command
In order to be able to distribute the projectors to different workers, you should be able to give one or more projector ids with each projector command.

Something has already been prepared and only needs to be implemented: https://github.com/patchlevel/event-sourcing/blob/2.1.x/src/Console/Command/ProjectionistCommand.php#L21

DavidBadura:
* restructure & rename projection classes
* fix #320
* add reactivate functionality
* add more tests
* add projectionist documentation
* fix projection documentation


src/Projection/ProjectorStore/ProjectorStateCollection.php
```diff
@@ -86,20 +86,27 @@ public function minProjectorPosition(): int
         return $min ?: 0;
     }
 
-    public function filterByProjectorStatus(ProjectorStatus $status): self
+    /**
+     * @param callable(ProjectorState $state):bool $callable
```
DanielBadura: `pure-callable` would remove the baseline entry i guess

src/Projection/ProjectorId.php
```diff
@@ -13,7 +20,7 @@ final class ProjectorId
 {
     public function __construct(
         private readonly string $name,
-        private readonly int $version
+        private readonly int $version = 1
```
DanielBadura: Do we really want this default behaviour here? I think we should force the user to think about it and keep track of it, no?

src/Console/Command/ProjectionistCommand.php
```diff
@@ -16,8 +25,50 @@ public function __construct(
...
+        $this
+            ->addOption(
+                'id',
+                null,
+                InputOption::VALUE_IS_ARRAY,
+                'The maximum number of runs this command should execute'
+            );
+    }
+
+    protected function projectorCriteria(InputInterface $input): ProjectorCriteria
```
DanielBadura: I'm not sure about passing here the whole `InputInterface` into it. We gain alot of power with this and possibilities but could it also backfire? WDYT?

DanielBadura (approved):

src/Projection/ProjectorId.php
DavidBadura: This is just a symptom of the problem i am having right now. In the documentation i noticed that it is difficult to explain what this ProjectorId is. And what a projection or projector is, or what the difference is.

I am going to revisit the naming and revisit the issue so that you only have to provide the ID if you are going to use it.

src/Console/Command/ProjectionistCommand.php
DavidBadura: Don't see the danger here

src/Pipeline/Target/ProjectionHandlerTarget.php
```diff
@@ -7,6 +7,9 @@
 use Patchlevel\EventSourcing\EventBus\Message;
 use Patchlevel\EventSourcing\Projection\ProjectionHandler;
 
+/**
+ * @deprecated ProjectorRepositoryTarget
```
DanielBadura:
```suggestion
 * @deprecated use ProjectorRepositoryTarget instead
```

src/Pipeline/Target/ProjectionTarget.php
```diff
@@ -9,6 +9,9 @@
 use Patchlevel\EventSourcing\Projection\MetadataAwareProjectionHandler;
 use Patchlevel\EventSourcing\Projection\Projection;
 
+/**
+ * @deprecated use ProjectorTarget
```
DanielBadura:
```suggestion
 * @deprecated use ProjectorTarget instead
```

src/Projection/Projection/ProjectionCollection.php
```diff
@@ -0,0 +1,134 @@
...
+
+use function array_filter;
+use function array_key_exists;
+use function array_values;
+use function count;
+
+/**
+ * @implements IteratorAggregate<int, Projection>
+ */
+final class ProjectionCollection implements Countable, IteratorAggregate
```
DanielBadura, line 19: So now you removed the `immutable` marker but then the API is weird to me. Since the collection is now returning a new instance everytime and not updating the state internally. I think this could be confusing for the user.

src/Projection/Projection/Store/DoctrineStore.php
```diff
@@ -0,0 +1,139 @@
...
+            'name' => $projectionId->name(),
+            'version' => $projectionId->version(),
+        ]);
+
+        if ($result === false) {
+            throw new ProjectionNotFound($projectionId);
+        }
+
+        return new Projection(
+            new ProjectionId($result['name'], $result['version']),
```
DanielBadura: Do we need to create a new ` ProjectionId`?

src/Projection/Projectionist/DefaultProjectionist.php
```diff
@@ -0,0 +1,307 @@
...
+                $this->logger?->info(sprintf('projection for "%s" prepared', $projection->id()->toString()));
+            } catch (Throwable $e) {
+                $this->logger?->error(sprintf('preparing error in "%s"', $projection->id()->toString()));
+                $this->logger?->error($e->getMessage());
+                $projection->error();
+                $this->projectorStore->save($projection);
+            }
+        }
+
+        $currentPosition = $projections->minProjectionPosition();
```
DanielBadura: Maybe a better naming for this method `getCurrentLowerstProjectionPosition` ?

DanielBadura (approved):

PR #340, opened 2022-12-13
add projectionist event bus wrapper

DavidBadura:
Then there is also the possibility to use the projectionist synchronously.


DavidBadura:
After further testing it turned out that it is not so easy after all, if not impossible. So I'm closing the PR because we decided not to pursue it further for now.

DanielBadura (approved):

PR #346, opened 2022-12-25
simplify normalizer usage

DavidBadura:
Before:

```php
final class DTO 
{
    #[Normalize(new NameNormalizer())]
    public Name $name

    #[Normalize(new DateTimeImmutableNormalizer(format: DateTimeImmutable::RFC3339_EXTENDED))]
    public DateTimeImmutable $date;

    #[Normalize(new EnumNormalizer(Status::class))]
    public Status $status;
}
```

After:

```php
final class DTO 
{
    #[NameNormalizer]
    public Name $name

    #[DateTimeImmutableNormalizer(format: DateTimeImmutable::RFC3339_EXTENDED)]
    public DateTimeImmutable $date;

    #[EnumNormalizer(Status::class)]
    public Status $status;
}
```


src/Attribute/Normalize.php
```diff
@@ -8,6 +8,9 @@
 use Patchlevel\EventSourcing\Serializer\Normalizer\ArrayNormalizer;
 use Patchlevel\EventSourcing\Serializer\Normalizer\Normalizer;
 
+/**
+ * @deprecated use the specific normalizer as attribute
```
DanielBadura: I would also maybe add a sentence about custom normalizers that they should add now the attribute to enable them as attributes

DanielBadura (approved):
Just a small addition maybe in the deprecation notice.

PR #375, opened 2023-05-03
fix missing default values for not nullable fields

DavidBadura:


DanielBadura (approved):

DanielBadura:
Should be merged in 2.1.x, no?

PR #436, opened 2023-12-13
store projection error objects

### LINKED ISSUE #397: A better exception processing for the projections
See how symfony messenger solves the problem

DavidBadura:
fix #397

### DIFF
--- a/src/Console/Command/ProjectionStatusCommand.php
+++ b/src/Console/Command/ProjectionStatusCommand.php
@@ -4,13 +4,17 @@ declare(strict_types=1);
 
 namespace Patchlevel\EventSourcing\Console\Command;
 
+use Patchlevel\EventSourcing\Console\InputHelper;
 use Patchlevel\EventSourcing\Console\OutputStyle;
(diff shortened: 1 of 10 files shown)


src/Console/Command/ProjectionStatusCommand.php
```diff
@@ -4,45 +4,97 @@
...
                 ],
-                [...$projections],
-            ),
+            ],
         );
 
+        $errorObject = $projection->errorObject();
+
+        if ($errorObject) {
+            dump($errorObject);
```
DanielBadura: should be removed

DanielBadura (changes requested):

src/Console/Command/ProjectionStatusCommand.php
DavidBadura: I want to dump the exception

DavidBadura: replaced

DanielBadura (approved):

PR #438, opened 2023-12-13
Fix README.md projectionist link

robinlehrmann:
It seams this link is outdated/wrong :)


DanielBadura:
Thanks, tbf all links are there wrong due to `latest` and we are already publishing our latest docs for `3.0`. The link would be right, if it would link to the version `2.3`: `https://patchlevel.github.io/event-sourcing-docs/2.3/projectionist/`. So i would think we need to fix all links from `latest` to `2.3` or not publish already our newest docs from our work in progress branch, wdyt @DavidBadura ?

DavidBadura:
The "latest" points to 3.0.0 is incorrect. I found the error in the workflow. When searching for the last version, he also looked at the prereleases and therefore already changed the alias to 3.0.0. I fixed that in the script (#439). I just have to see how I can change the alias by hand.

PR #454, opened 2024-01-05
Add ProjectorId as value object

DanielBadura:


src/Projection/Projectionist/DefaultProjectionist.php
```diff
@@ -348,7 +348,8 @@ public function projections(): ProjectionCollection
         $projectors = $this->projectors();
 
         foreach ($projectors as $projector) {
-            $projectionId = $this->projectorResolver->projectionId($projector);
+            $projectorId = $this->projectorResolver->projectorId($projector);
+            $projectionId = new ProjectionId($projectorId->name(), $projectorId->version());
```
DavidBadura: I think this could be made a little nicer like this:

```php
$projectionId = $projectorId->toProjectionId();
```

and handle it internally

DavidBadura (approved):

PR #465, opened 2024-01-11
refactor default event bus & drop symfony event bus

DavidBadura:


src/Attribute/Subscribe.php
```diff
@@ -9,7 +9,9 @@
 #[Attribute(Attribute::TARGET_METHOD | Attribute::IS_REPEATABLE)]
 final class Subscribe
 {
-    /** @param class-string $eventClass */
+    public const ALL = '*';
+
+    /** @param class-string|'*' $eventClass */
```
DanielBadura: maybe `class-string|self::ALL` ?

src/EventBus/AttributeListenerProvider.php
```diff
@@ -0,0 +1,58 @@
...
+                    $this->subscribeMethods[$eventClass][] = new ListenerDescriptor(
+                        $listener->{$method->getName()}(...),
+                    );
+                }
+            }
+        }
+
+        return array_merge(
+            $this->subscribeMethods[$event::class] ?? [],
+            $this->subscribeMethods['*'] ?? [],
```
DanielBadura: use here the public const: `Subscribe::ALL` ?

src/EventBus/ListenerDescriptor.php
```diff
@@ -0,0 +1,53 @@
...
+    private readonly Closure $callable;
+    private readonly string $name;
+
+    public function __construct(callable $callable)
+    {
+        $callable = $callable(...);
+
+        $this->callable = $callable;
+
+        $r = new ReflectionFunction($callable);
```
DanielBadura: `$r` => `$reflectionFunction`

src/Metadata/Projector/AttributeProjectorMetadataFactory.php
```diff
@@ -47,6 +47,13 @@ public function metadata(string $projector): ProjectorMetadata
                 $instance = $attribute->newInstance();
                 $eventClass = $instance->eventClass;
 
+                if ($eventClass === '*') {
```
DanielBadura: We could also use here the public const `Subscribe::ALL`

src/WatchServer/WatchListener.php
```diff
@@ -4,16 +4,17 @@
...
-final class WatchListener implements Listener
+final class WatchListener
 {
     public function __construct(
-        private WatchServerClient $client,
+        private readonly WatchServerClient $client,
     ) {
     }
 
+    #[Subscribe('*')]
```
DanielBadura: We could even use here the const, no?

DanielBadura (approved):

PR #477, opened 2024-02-01
Add deptrac

DanielBadura:


composer.json
```diff
@@ -45,6 +45,7 @@
     "phpstan/phpstan": "^1.10.49",
     "phpunit/phpunit": "^10.5.2",
     "psalm/plugin-phpunit": "^0.18.4",
+    "qossmic/deptrac": "dev-main#5dd4bbc507fb6b36468c8de889bdd2d586e4a580",
```
DanielBadura: tbd, maybe move this to tools due to instability? 

deptrac-shim is to be removed and i think the normal deptrac release will be the shim version.

deptrac.yaml
```diff
@@ -0,0 +1,139 @@
...
+      - Aggregate
+      - Attribute
+      - Metadata
+      - Serializer
+    Lock:
+      - Schema
+    Metadata:
+      - Aggregate
+      - Attribute
+      - Projection
```
DanielBadura: i think, we should move `Metadata` into the respecting domains. We can see here that we have cross referencing.

deptrac.yaml
```diff
@@ -0,0 +1,139 @@
...
+    - name: WatchServer
+      collectors:
+        - type: directory
+          value: src/WatchServer/.*
+
+  ruleset:
+    Aggregate:
+      - Attribute
+      - Metadata
+    Attribute:
```
DanielBadura, line 92: Right now, we also have a dependency into `Projection` from `Attribute` due to the default values. I dont think that this is a real problem, but if I open this here up, then maybe we could get some problems later on. WDYT?

composer.json
DanielBadura: Here some refs:
* https://github.com/qossmic/deptrac/issues/1345
* https://github.com/qossmic/deptrac/issues/1243
* https://github.com/qossmic/deptrac-shim/pull/11

DavidBadura: Yes, we should move it to tools. 👍

deptrac.yaml
DavidBadura: I don't think we should split it. I think it's good that it's in this `Metadata` namespace. This is exactly the same special case as `Console` and `Attribute`.

deptrac.yaml
DavidBadura: This will be removed soon.

deptrac.yaml
DanielBadura: I updated the config so it is split up for deptrac to have a better check on these layers. Maybe have a look on that.

I'm still thinking we should move it tho

DavidBadura (approved):

PR #494, opened 2024-02-06
Update Message with headers

DanielBadura:


src/EventBus/Message.php
```diff
@@ -17,8 +17,18 @@
...
  *  }
+ *
+ * @phpstan-type Headers = array{
+ *      aggregateName?: string,
+ *      aggregateId?: string,
+ *      playhead?: positive-int,
+ *      recordedOn?: DateTimeImmutable,
+ *      newStreamStart?: bool,
+ *      archived?: bool
+ *   }
```
DanielBadura, line 30: I added a extra type for PHPStan for now since unsealed array as a feature does not exist there. PHPStan only fixed syntax crashes for now with the `...` but type resolution is less useful as with this type we have here imho.

DavidBadura (approved):

PR #496, opened 2024-02-10
Add Message header interface for more decoupling and type safety

DanielBadura:


src/EventBus/Message.php
```diff
@@ -59,7 +42,10 @@ public static function create(object $event): self
         return new self($event);
     }
 
-    /** @param Headers $headers */
+    /**
+     * @template H of Header
+     * @param array<class-string<H>, H> $headers
```
DavidBadura: We shouldn't reveal the internal structure to the outside.

```
@param list<Header> $headers
```

src/EventBus/Message.php
```diff
@@ -186,43 +75,32 @@ public function header(string $name): mixed
...
     {
         return $this->headers;
     }
 
-    /** @param Headers $headers */
+    /**
+     * @template H of Header
+     * @param array<class-string<H>, H> $headers
+     */
     public function withHeaders(array $headers): self
```
DavidBadura: We shouldn't reveal the internal structure to the outside.

```php
/**
 * @param list<Header> $headers
 */
public function withHeaders(array $headers): self
{
    $message = clone $this;
    foreach($headers as $header) {
      $message[$header::class] = $header;
    }
    
    return $message;
}
```

src/EventBus/Message.php
```diff
@@ -186,43 +75,32 @@ public function header(string $name): mixed
...
-        $message->headers[$name] = $value;
+        $message->headers[$header::class] = $header;
 
         return $message;
     }
 
-    /** @return Headers */
+    /**
+     * @template H of Header
+     * @return array<class-string<H>, H>
```
DavidBadura:
```suggestion
     * @return list<Header>
```

src/Outbox/OutboxHeader.php
```diff
@@ -0,0 +1,18 @@
...
+declare(strict_types=1);
+
+namespace Patchlevel\EventSourcing\Outbox;
+
+use Patchlevel\EventSourcing\EventBus\Header;
+
+/**
+ * @psalm-immutable
+ */
+final class OutboxHeader implements Header
```
DavidBadura:
```suggestion
final class MessageIdentifierHeader implements Header
```

src/EventBus/Message.php
```diff
@@ -71,112 +57,15 @@ public function event(): object
...
-        return $value;
-    }
-
-    public function withArchived(bool $value): self
-    {
-        return $this->withHeader(self::HEADER_ARCHIVED, $value);
-    }
-
-    /** @throws HeaderNotFound */
     public function header(string $name): mixed
```
DavidBadura:
```suggestion
    public function header(string $name): Header
```

Maybe we should generally adjust the API to return null if the header doesn't exist.
This is much more developer friendly.

```php
    public function header(string $name): Header|null
```

From this:

```php
            try {
                $newStreamStartHeader = $message->header(NewStreamStartHeader::class);
            } catch (HeaderNotFound) {
                continue;
            }

            if (!$newStreamStartHeader->newStreamStart) {
                continue;
            }

            $lastMessageWithNewStreamStart = $message;
        }
```

We can do this:

```php
            if (!$message->header(NewStreamStartHeader::class)?->newStreamStart) {
                continue;
            }
        }
```

DavidBadura (changes requested):
That's kind of how I thought it too 👍

src/EventBus/Message.php
DavidBadura: we should use `iterable` here

```php
/**
 * @param iterable<Header> $headers
 */
public function withHeaders(iterable $headers): self
{
    $message = clone $this;
    foreach($headers as $header) {
      $message[$header::class] = $header;
    }
    
    return $message;
}
```

src/EventBus/Message.php
DavidBadura:
```suggestion
     * @param iterable<Header> $headers
```

DanielBadura:
Do we even need the `SerializedHeader` dto?

src/Attribute/HeaderIdentifier.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Patchlevel\EventSourcing\Attribute;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_CLASS)]
+final class HeaderIdentifier
```
DavidBadura: If we remove the "Header" meta interface, we can rename it back to Header, which I think is better.

src/EventBus/Header.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Patchlevel\EventSourcing\EventBus;
+
+/** @psalm-immutable */
+interface Header
```
DavidBadura: Can we remove this "meta" interface? We have the attribute already.

src/EventBus/Message.php
```diff
@@ -38,7 +22,7 @@ final class Message
     public const HEADER_ARCHIVED = 'archived';
     public const HEADER_NEW_STREAM_START = 'newStreamStart';
```
DavidBadura: we should remove the constants

src/EventBus/Serializer/HeadersSerializer.php
```diff
@@ -0,0 +1,25 @@
...
+
+use Patchlevel\EventSourcing\EventBus\Header;
+
+interface HeadersSerializer
+{
+    /**
+     * @param list<Header>         $headers
+     * @param array<string, mixed> $options
+     *
+     * @return array<SerializedHeader>
```
DavidBadura: can we change the interface to `array<string, array<string, mixed>>`? So we can rid of `SerializedHeader `?

src/EventBus/Serializer/SerializedHeader.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Patchlevel\EventSourcing\EventBus\Serializer;
+
+final class SerializedHeader
```
DavidBadura: I think this class is not needed

tests/Benchmark/SimpleSetupBench.php
```diff
@@ -49,6 +50,10 @@ public function setUp(): void
         $this->store = new DoctrineDbalStore(
             $connection,
             DefaultEventSerializer::createFromPaths([__DIR__ . '/BasicImplementation/Events']),
+            DefaultHeadersSerializer::createFromPaths([
```
DavidBadura, line 40: We should add the default headers internally, so the developer no need to know, that he need a specific path. And we should do this without a path: we should add our headers directly. auto-discovery should only used in user space.

DavidBadura (changes requested):
Nice work 👍 I still have a few wishes :D

tests/Benchmark/SimpleSetupBench.php
DanielBadura: Hm, I'm not sure about that. Due to that it's more masked for the user that this is happening and if he implements his own version of the `HeadersSerializer` he needs to know this internal logic in our `DefaultHeadersSerializer`.

But if we we do this then i think we should keep it as auto discover due two reasons:
1. We cannot forget to add headers then there
2. No need for extra logic

src/EventBus/Serializer/HeadersSerializer.php
DanielBadura: I can for sure check that

src/EventBus/Header.php
DanielBadura: So i have mixed feelings about this: we dont need that anymore thats true. But what about the psalm immutable annotation? Can we maybe add this information via our psalm plugin (also should do this then for events)? As i think this is needed due to the nature of `Message` being immutable.

tests/Benchmark/SimpleSetupBench.php
DavidBadura: 1) I don't think it's a relevant point that you forget it, since we'll hopefully cover it with tests so that everything fits.

2) We can push the logic further down to `MessageHeaderRegistry`.

```php
class MessageHeaderRegistry
{
    // ...
    
    /** @param array<string, class-string<Header>> $headerNameToClassMap */
    public static function withInternalHeaders(array $headerNameToClassMap): self
    {
        $internalHeaders = [
            'aggregate' => AggregateHeader::class,
            // ...
        ];
    
        return new self($headerNameToClassMap + $internalHeaders);
    }
}
```

This means it cannot be forgotten in the serializer.

To the point that no extra logic is necessary: It's an array merge, there isn't that much extra logic. On the other hand, we have an IO search in our code, which is not necessary because we know exactly where the classes are.

src/EventBus/Header.php
DavidBadura: I'm generally against meta interfaces because an interface should offer something, otherwise you can leave it out.

The attribute `@psalm-immutable` doesn't help either: https://psalm.dev/r/8427a00052

I also don't think that we should educate the user with something like this, but rather write it in the documentation, which explains how you can add a header yourself.

DanielBadura: The example is more like this: https://psalm.dev/r/290007b891
Still not sure if it would catch something dirty doing by the user. 

I dont think this should be framed as a bad "education". This could lead to bugs if not all methods/calls are pure due e.g. serialization. 

But as i mentioned before, maybe this should be placed then in the psalm plugin, IF we want/need this, since we dont need this interface for anything else.

tests/Benchmark/SimpleSetupBench.php
DanielBadura: Then we are duplicating the naming code 🤔 
So i think, if we move it, it should land in `AttributeMessageHeaderRegistryFactory` but still feels weird.
```php
        $classes = (new ClassFinder())->findClassNames($paths);
        $classes = array_merge([AggregateHeader::class, ...], $classes);
```

src/EventBus/Header.php
DanielBadura: I removed it: as i said, more suitable in plugin i think.

tests/Benchmark/SimpleSetupBench.php
DavidBadura: > Then we are duplicating the naming code 🤔

I do not understand that.

For me, `MessageHeaderRegistry` would be the more correct place. I wouldn't put it any further down.

DavidBadura: If you mean the attributes. I would remove these from the internal headers because the attributes are only used for configuration. And since we define the internal headers differently, they have no meaning there. The attribute is for user space.

PR #500, opened 2024-02-17
"from now" mode for projections

DavidBadura:


src/Projection/Projectionist/DefaultProjectionist.php
```diff
@@ -593,4 +590,92 @@ private function handleRetryProjections(ProjectionCollection $projections): void
             );
         }
     }
+
+    private function handleFromNowProjections(ProjectionCollection $projections): void
```
DanielBadura, line 594:
```suggestion
    private function activateFromNowProjectionsWithLatestIndex(ProjectionCollection $projections): void
```
Maybe to be more clear?

src/Projection/Projectionist/DefaultProjectionist.php
```diff
@@ -593,4 +590,92 @@ private function handleRetryProjections(ProjectionCollection $projections): void
...
+
+            if (!$projector) {
+                continue;
+            }
+
+            $metadata = $this->metadataFactory->metadata($projector::class);
+
+            if (!$metadata->fromNow) {
+                continue;
+            }
```
DanielBadura, line 609: maybe extract this filtering in a private method to increase readability and reduce the complexity in this method?

DanielBadura (approved):

src/Projection/Projectionist/DefaultProjectionist.php
DavidBadura: I will do this in a following PR as I have started building the Projectionist differently.

src/Projection/Projectionist/DefaultProjectionist.php
DavidBadura: I will do this in a following PR as I have started building the Projectionist differently.

PR #503, opened 2024-02-18
overhaul projectionist

### LINKED ISSUE #321: projector tag filter option for all projectionist command
In order to be able to distribute the projectors to different workers, you should be able to give each projector one or more tags. Each projectionist command should then be able to get a filter for one or more tags.

Another improvement to the ticket #320

DavidBadura:
I rewrote the Projectionist core so that it no longer needs an external lock. The "FOR UPDATE" statement is used to lock the rows when reading as long as the transaction exists. If the row is locked, no other process can read it. This means there is no risk of events being processed twice. In addition, only the rows that were affected by the “fetch” are locked. This means that the projections and thus Projectionist can also be run in groups.

Features:

* Implement row lock
* Allow grouping of projections
* Add new run mode named "once"
* Cleanup code
* Improve performance

resolve: https://github.com/patchlevel/event-sourcing/issues/321


src/Projection/Projection/Store/DoctrineStore.php
```diff
@@ -55,94 +59,111 @@ public function get(string $projectionId): Projection
...
     {
-        $sql = $this->connection->createQueryBuilder()
+        $qb = $this->connection->createQueryBuilder()
             ->select('*')
-            ->from($this->projectionTable)
-            ->getSQL();
+            ->from($this->projectionTable);
+
+        if (!$this->connection->getDatabasePlatform() instanceof SQLitePlatform) {
+            $qb->forUpdate();
```
DanielBadura, line 73: First, i thought about an interface for providing locking and using a "dbal" implementation here, but maybe this is too much? 

Second, checking on the platform here is maybe a bit brittle, no? DBAL could easily add a new platform which also does not support this feature and then we would not catch this with our check here unless updated.

src/Projection/Projection/Store/InMemoryStore.php
```diff
@@ -25,22 +37,68 @@ public function get(string $projectionId): Projection
...
+
+    public function remove(Projection $projection): void
+    {
+        unset($this->projections[$projection->id()]);
+    }
+
+    public function transactional(Closure $closure): void
+    {
+        $closure();
     }
```
DanielBadura: This is here not really transactional. We would first need to answer the question: Should this class be used outside of tests? And if yes, which could be done i think, then we should implement this to match the behaviour which the user expects it like.

src/Projection/Projection/Store/DoctrineStore.php
DavidBadura: > First, i thought about an interface for providing locking and using a "dbal" implementation here, but maybe this is too much?

I thought about it too, but there were too many if statements at the time. But now I could do it because I abstracted the part well in the projectionist.

> Second, checking on the platform here is maybe a bit brittle, no? DBAL could easily add a new platform which also does not support this feature and then we would not catch this with our check here unless updated.

It doesn't really matter whether DBAL adds a new platform or whether SQLite implements the feature in the future. We listed the supported DB versions in the Readme and expressly said that we only support them. These are also tested in the CI.

I can't solve it any other way. There is no "supportForUpdate" method.

src/Projection/Projection/Store/InMemoryStore.php
DavidBadura: I solved the problem with `TransactionalStore` interface.

DanielBadura (approved):

PR #510, opened 2024-02-25
upgrade hydrator & use new auto detect feature

DavidBadura:
needs https://github.com/patchlevel/hydrator/pull/37


DanielBadura:
I think we also should update the BenchmarkTest Fixtures to reflect the performance changes - which I expect to be low to none.

DavidBadura:
Then the benchmark should be added to the hydrator, what do you think?

DavidBadura:
I have now rewrite everything to the new system, including in the benchmark test.

DanielBadura (approved):

PR #511, opened 2024-02-26
allow to pause projections

DavidBadura:


src/Projection/Projectionist/DefaultProjectionist.php
```diff
@@ -473,6 +478,48 @@ function (array $projections): void {
         );
     }
 
+    public function pause(ProjectionistCriteria|null $criteria = null): void
+    {
+        $criteria ??= new ProjectionistCriteria();
+
+        $this->discoverNewProjections();
```
DanielBadura, line 485: should this happen if we want to pause a projection? Seems weird to me 🤔

DanielBadura (approved):
besides my question looking great

src/Projection/Projectionist/DefaultProjectionist.php
DavidBadura: The logic, discoverNewProjections, is executed on every command. Also access when reading.

If this is done differently here, it will be confusing. We should rather question when exactly this logic happens.

PR #521, opened 2024-03-04
try guid for aggregate id

DavidBadura:


DavidBadura:
This makes no difference, so it is closed.

Issue #481, opened 2024-02-01
Prepare docs for 3.0.0


DavidBadura:
- [x] readme https://github.com/patchlevel/event-sourcing/pull/440
- [x] aggregate https://github.com/patchlevel/event-sourcing/pull/483
- [x] aggregate id https://github.com/patchlevel/event-sourcing/pull/483
- [x] cli https://github.com/patchlevel/event-sourcing/pull/505
- [x] clock https://github.com/patchlevel/event-sourcing/pull/483
- [x] message https://github.com/patchlevel/event-sourcing/pull/564
- [x] event bus https://github.com/patchlevel/event-sourcing/pull/483
- [x] events https://github.com/patchlevel/event-sourcing/pull/483
- [x] getting started https://github.com/patchlevel/event-sourcing/pull/440
- [x] index https://github.com/patchlevel/event-sourcing/pull/440
- [x] message decorator https://github.com/patchlevel/event-sourcing/pull/483
- [x] normalizer https://github.com/patchlevel/event-sourcing/pull/510
- [x] projection https://github.com/patchlevel/event-sourcing/pull/503
- [x] repository https://github.com/patchlevel/event-sourcing/pull/507
- [x] snapshots https://github.com/patchlevel/event-sourcing/pull/507
- [x] splitstream https://github.com/patchlevel/event-sourcing/pull/507
- [x] store https://github.com/patchlevel/event-sourcing/pull/505
- [x] tests https://github.com/patchlevel/event-sourcing/pull/507
- [x] upcasting https://github.com/patchlevel/event-sourcing/pull/507

PR #523, opened 2024-03-07
remove outbox store

DavidBadura:
The outbox store is removed because it simply duplicated the event store. Instead, you can use a subscriber that does exactly the same thing:

```php
<?php

namespace App\Share\Infrastructure\EventSourcing;

use Patchlevel\EventSourcing\Attribute\Projector;
use Patchlevel\EventSourcing\EventBus\Message;
use Patchlevel\EventSourcing\Projection\Projection\RunMode;

#[Projector('outbox', runMode: RunMode::FromNow)]
class OutboxProcessor
{
    public function __construct(
        private readonly EventBus $eventBus,
    ) {
    }

    #[Subscribe('*')]
    public function publish(Message $message): void
    {
        $this->eventBus->dispatch($message);
    }
}
```


DanielBadura (approved):
Do we need to add more context in the docs? Still referencing outbox and why we dont use it (anymore)?

DavidBadura:
I'll rewrite this with the projection and pipeline, as soon as the namming etc. is correct.
https://github.com/patchlevel/event-sourcing/issues/481

PR #526, opened 2024-03-08
rename projection into subscription

### LINKED ISSUE #513: rename projectionist / projection engine
Since projectionist has now become more powerful and can therefore not only be used for projections, the system should be renamed.

The following terms are used:

* **Projector:** Is the class that subscribes to Events and executes its action
* **Projection:** Is an object that is stored in the ProjectionStore and contains all information such as position and status.
* **Projectionist:** Is the engine that carries out the respective actions using Projectors (Executer) and Projection (State).

1) We are talking about subscriptions. There are two of them under categories like Projector (Projections) and Processor (Listener).

* Projector -> Subscriber (or in user space Subscriber/Projector/Processor)
* Projection -> Subscription?
* Projectionist -> Engine?

2) We're talking about processors. And projection/projector is a special form of it.

* Projector -> Processor (or in user space Projector/Processor)
* Projection -> Process?
* Projectionist -> ProcessorManager

DavidBadura:
fix #513


DanielBadura (approved):
I think we should still use projections and processors in the docs for clarificaiton. But this is fine for now i think 👍

PR #527, opened 2024-03-08
Fix customHeader method to actually check for the header existing

gzumba:
Fixes the customHeader method implementation which was simply broken.


DavidBadura:
Good catch! can you fix the coding style? Thanks!

DavidBadura (approved):

DavidBadura:
Thank you, is fixed in 2.3.2 https://github.com/patchlevel/event-sourcing/releases/tag/2.3.2

PR #538, opened 2024-03-13
move trace feature into debug namespace

DavidBadura:


DanielBadura (approved):
LGTM. This need changes in the deptrac config which now landed. Could you update it please?

PR #546, opened 2024-03-19
add event payload cryptographer for personal data

DavidBadura:
I took this blog post as inspiration and built a prototype:

https://www.eventstore.com/blog/protecting-sensitive-data-in-event-sourced-systems-with-crypto-shredding-1

This is what an event looks like when you use personal data:

```php
<?php
use Patchlevel\EventSourcing\Attribute\DataSubjectId;
use Patchlevel\EventSourcing\Attribute\Event;
use Patchlevel\EventSourcing\Attribute\PersonalData;
use Patchlevel\EventSourcing\Serializer\Normalizer\IdNormalizer;

#[Event('profile.created')]
final class ProfileCreated
{
    public function __construct(
        #[IdNormalizer]
        #[DataSubjectId]
        public ProfileId $profileId,
        #[PersonalData(fallback: 'unknown')]
        public string $name,
    ) {
    }
}
```

And you can use a processor to remove personal data:

```php
use Patchlevel\EventSourcing\Attribute\Processor;
use Patchlevel\EventSourcing\Cryptography\Store\CipherKeyStore;
use Patchlevel\EventSourcing\Message\Message;

#[Processor('delete_personal_data')]
final class DeletePersonalDataProcessor
{
    public function __construct(
        private readonly CipherKeyStore $cipherKeyStore,
    ) {
    }

    #[Subscribe(UserHasRequestedDeletion::class)]
    public function handleUserHasRequestedDeletion(Message $message): void
    {
        $event = $message->event();

        $this->cipherKeyStore->remove($event->profileId->toString());
    }
}
```


src/Attribute/DataSubjectId.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Patchlevel\EventSourcing\Attribute;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_PROPERTY)]
+final class DataSubjectId
```
DanielBadura, line 10: This naming coming from the article you mentioned at the PR. But i could find in the article *why* it is named like it is. For me the naming is not really making sense in the context. Do I miss something?

src/Cryptography/Store/InMemoryCipherKeyStore.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Patchlevel\EventSourcing\Cryptography\Store;
+
+use Patchlevel\EventSourcing\Cryptography\Cipher\CipherKey;
+
+final class InMemoryCipherKeyStore implements CipherKeyStore
```
DanielBadura, line 9: Im not sure if this makes sense to have in our `src` dir. For testing purposes i can understand that, but on production systems?

src/Cryptography/Store/DoctrineCipherKeyStore.php
```diff
@@ -0,0 +1,105 @@
...
+ *     subject_id: non-empty-string,
+ *     crypto_key: non-empty-string,
+ *     crypto_method: non-empty-string,
+ *     crypto_iv: non-empty-string
+ * }
+ */
+final class DoctrineCipherKeyStore implements CipherKeyStore, DoctrineSchemaConfigurator
+{
+    /** @var array<string, CipherKey> */
+    private array $keys = [];
```
DanielBadura: maybe name it keyCache or sth like that?

src/Attribute/DataSubjectId.php
DavidBadura: The point is that every person-related data must be assigned to a subject. A subject can be a person. To refer to this subject, you need to use a identifier. So we have the name `SubjectId`. And I think `Data` is a prefix to scope it and understand that this is about a subject ID related to the data.

src/Cryptography/Store/InMemoryCipherKeyStore.php
DavidBadura: Classes that are used (only) for tests and offered to the developer also belong in the src folder.
I definitely want to test this class too. This is not a stub or mock, but a valid implementation of the interface.

DanielBadura (approved):

PR #548, opened 2024-03-21
 docs: format & fix php errors

DavidBadura:


.github/workflows/docs-check.yml
```diff
@@ -0,0 +1,50 @@
+# https://help.github.com/en/categories/automating-your-workflow-with-github-actions
+
+name: "Check Docs"
+
+on:
+    pull_request:
+    push:
+        branches:
+            - "[0-9]+.[0-9]+.x"
+            - "renovate/*"
```
DanielBadura: Do we need this on renovate pushes?

.github/workflows/docs-check.yml
```diff
@@ -0,0 +1,50 @@
...
+                dependency-versions: ${{ matrix.dependencies }}
+
+            - name: "extract php code"
+              run: "bin/docs-extract-php-code"
+
+            - name: "lint php"
+              run: "php -l docs_php/*.php"
+
+            - name: "docs code style"
+              run: "vendor/bin/phpcbf docs_php --exclude=SlevomatCodingStandard.TypeHints.DeclareStrictTypes"
```
DanielBadura, line 49: i like this solution perse: i wonder how the checks from othe SA tools will change due to the "missing" `declare(strict_types=1);`. Should be minor i think, but still WDYT?

composer.json
```diff
@@ -38,6 +38,7 @@
     "cspray/phinal": "^2.0.0",
     "doctrine/orm": "^2.18.0|^3.0.0",
     "infection/infection": "^0.27.0",
+    "league/commonmark": "^2.4",
```
DanielBadura, line 41: maybe move to tools?

composer.json
```diff
@@ -48,7 +49,8 @@
     "roave/infection-static-analysis-plugin": "^1.34.0",
     "symfony/messenger": "^5.4.31|^6.4.0|^7.0.1",
     "symfony/var-dumper": "^5.4.29|^6.4.0|^7.0.0",
-    "vimeo/psalm": "^5.17.0"
+    "vimeo/psalm": "^5.17.0",
+    "wnx/commonmark-markdown-renderer": "^1.4"
```
DanielBadura, line 53: maybe move to tools?

DanielBadura (approved):

.github/workflows/docs-check.yml
DavidBadura: I don't know if it behaves differently, but we can't change anything about it, otherwise we would always be formatting because we always remove that. But we already have 500 psalm errors that need to be fixed/ignored :D

composer.json
DavidBadura: We could try it out and see if it works. But I don't want to do that here anymore. I had to start over 3 times because I had an error in the script. I want to save the status for now :D

PR #549, opened 2024-03-21
Add infection on diff

DanielBadura:


.github/workflows/mutation-tests.yml
```diff
@@ -44,3 +44,38 @@ jobs:
...
+                  extensions: pdo_sqlite
+
+            - uses: ramsey/composer-install@2.2.0
+              with:
+                  dependency-versions: ${{ matrix.dependencies }}
+
+            - name: "Infection"
+              run: "vendor/bin/roave-infection-static-analysis-plugin --git-diff-lines --git-diff-base=origin/$GITHUB_BASE_REF --ignore-msi-with-no-mutations --only-covered --min-msi=80 --min-covered-msi=100"
+              env:
+                  STRYKER_DASHBOARD_API_KEY: ${{ secrets.STRYKER_DASHBOARD_API_KEY }}
```
DavidBadura: I do not think we want to store the value.

.github/workflows/mutation-tests.yml
```diff
@@ -44,3 +44,38 @@ jobs:
               run: "vendor/bin/roave-infection-static-analysis-plugin"
               env:
                   STRYKER_DASHBOARD_API_KEY: ${{ secrets.STRYKER_DASHBOARD_API_KEY }}
+
+    mutation-tests-diff:
```
DavidBadura: This should only run in PR.

DavidBadura (approved):

PR #563, opened 2024-04-03
Bench assertion

DanielBadura:


.github/workflows/benchmark.yml
```diff
@@ -67,7 +67,7 @@ jobs:
           dependency-versions: ${{ matrix.dependencies }}
 
       - name: "phpbench diff"
-        run: "vendor/bin/phpbench run tests/Benchmark --progress=none --report=diff --ref=base > bench.txt"
+        run: 'vendor/bin/phpbench run tests/Benchmark --progress=none --report=diff --ref=base --assert="mode(variant.time.avg) <= mode(baseline.time.avg) +/- 5%"  --assert="mode(variant.mem.peak) <= mode(baseline.mem.peak) +/- 5%" > bench.txt'
 
       - name: "Get Bench Result"
```
DavidBadura, line 73: we should add `if: always()` so we can see the result.

Or maybe: `if: ${{ success() || steps.diff.conclusion == 'failure' }}` and add in the step before the id `diff`

https://docs.github.com/en/actions/learn-github-actions/expressions#example-of-failure-with-conditions

DavidBadura: It would be best if we could react to status code 2.

DavidBadura (approved):

PR #573, opened 2024-04-12
add catch up subscription engine

DavidBadura:
waiting for #575 and #574

### DIFF
--- a/src/Subscription/Engine/CatchUpSubscriptionEngine.php
+++ b/src/Subscription/Engine/CatchUpSubscriptionEngine.php
@@ -1,0 +1,107 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Patchlevel\EventSourcing\Subscription\Engine;
+
+use Patchlevel\EventSourcing\Subscription\Subscription;
+
+use function array_merge;
+
+use const PHP_INT_MAX;
+
+final class CatchUpSubscriptionEngine implements SubscriptionEngine
+{
+    public function __construct(
+        private readonly SubscriptionEngine $parent,
+        private readonly int|null $limit = null,
+    ) {
+    }
+
+    public function setup(SubscriptionEngineCriteria|null $criteria = null, bool $skipBooting = false): Result
+    {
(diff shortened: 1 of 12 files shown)


src/Subscription/Engine/CatchUpSubscriptionEngine.php
```diff
@@ -0,0 +1,97 @@
...
+
+use Patchlevel\EventSourcing\Subscription\Subscription;
+
+use const PHP_INT_MAX;
+
+final class CatchUpSubscriptionEngine implements SubscriptionEngine
+{
+    public function __construct(
+        private readonly SubscriptionEngine $parent,
+        private readonly int|null $limit = null,
```
DanielBadura, line 17: Maybe instead of `null` add here `INT_MAX` as default value?

src/Subscription/Engine/CatchUpSubscriptionEngine.php
```diff
@@ -0,0 +1,97 @@
...
+
+    public function boot(SubscriptionEngineCriteria|null $criteria = null, int|null $limit = null): ProcessedResult
+    {
+        $mergedResult = new ProcessedResult(0);
+
+        $catchupLimit = $this->limit ?? PHP_INT_MAX;
+
+        for ($i = 0; $i < $catchupLimit; $i++) {
+            $result = $this->parent->boot($criteria, $limit);
+            $mergedResult = $this->mergeResult($mergedResult, $result);
```
DanielBadura: Maybe instead of merging in every iteration, merge after the loop?

src/Subscription/Engine/CatchUpSubscriptionEngine.php
DavidBadura: I had that before and changed it to `null`. The reason is I would also have to set it as a default value in Bundle. And if we decide to add a third default parameter (e.g. logger) and the developer only wants to set this, he must also set this constant.

Regarding the constant itself: This is an implementation detail. I took the constant because it works internally with a `for` loop. If the code is rewritten internally, it may also be that the constant is no longer needed.

PR #581, opened 2024-04-18
make aggregate id type configurable between uuid and string

DavidBadura:
In some cases it can make sense and because of BC to event sourcing v2, the aggregate id type must be configurable in the store. You can choose between `string` and `uuid`. Default will now be `uuid`.

### DIFF
--- a/docs/pages/aggregate_id.md
+++ b/docs/pages/aggregate_id.md
@@ -7,6 +7,12 @@ since only an aggregate-wide unique string is expected in the store.
 
 This library provides you with a few options for generating the id.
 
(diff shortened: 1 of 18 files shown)


DanielBadura (approved):
Maybe it makes sense to provide the strings as consts, but maybe its also not really worth it

DavidBadura:
I don't think it's necessary, we have a psalm for that.

DanielBadura:
But the user is not forced to use psalm. But i think PHPStorm nowadays also show the possibilities. So i also dont think its needed.

PR #591, opened 2024-04-23
Add migration path from MultiTableStore to SingelTableStore

DanielBadura:
This is a pre-requirement if users want to upgrade to v3 and are using the `MultiTableStore`.


DanielBadura:
Not sure if we should target 2.2.x instead and make just for the docs a release?

DavidBadura:
Doc changes do not have to be released and should not trigger a release.

DavidBadura (approved):

PR #607, opened 2024-06-05
use transaction id to not lose events

DavidBadura:
try to fix https://github.com/patchlevel/event-sourcing/issues/615

Implementation of:

* https://event-driven.io/en/ordering_in_postgres_outbox/
* https://github.com/eugene-khyst/postgresql-event-sourcing


DavidBadura:
replaced by https://github.com/patchlevel/event-sourcing/pull/613

PR #608, opened 2024-06-17
Add support for iterators when configuring ArgumentResolvers

gzumba:
This is required for the bundle to be able to use tagged_iterator for auto configuring ArgumentResolvers

### DIFF
--- a/src/Subscription/Subscriber/MetadataSubscriberAccessorRepository.php
+++ b/src/Subscription/Subscriber/MetadataSubscriberAccessorRepository.php
@@ -26,7 +26,7 @@ final class MetadataSubscriberAccessorRepository implements SubscriberAccessorRe
 
     /**
      * @param iterable<object>       $subscribers
-     * @param list<ArgumentResolver> $argumentResolvers
+     * @param iterable<ArgumentResolver>|list<ArgumentResolver> $argumentResolvers
      */
     public function __construct(
         private readonly iterable $subscribers,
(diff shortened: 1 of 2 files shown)


src/Subscription/Subscriber/MetadataSubscriberAccessorRepository.php
```diff
@@ -31,10 +31,11 @@ final class MetadataSubscriberAccessorRepository implements SubscriberAccessorRe
     public function __construct(
         private readonly iterable $subscribers,
         private readonly SubscriberMetadataFactory $metadataFactory = new AttributeSubscriberMetadataFactory(),
-        array $argumentResolvers = [],
+        iterable $argumentResolvers = [],
```
DanielBadura, line 36: We would also need to update the type for `$argumentResolvers` in the doc-block.
```php

    /**
     * @param iterable<object> $subscribers
     * @param iterable<ArgumentResolver>|list<ArgumentResolver> $argumentResolvers
     */
```

src/Subscription/Subscriber/MetadataSubscriberAccessorRepository.php
```diff
@@ -31,10 +31,11 @@ final class MetadataSubscriberAccessorRepository implements SubscriberAccessorRe
...
     ) {
         $this->argumentResolvers = array_merge(
-            $argumentResolvers,
+        // the check for array is required before PHP 8.2
+            is_array($argumentResolvers) ? $argumentResolvers : iterator_to_array($argumentResolvers),
```
DanielBadura:
```suggestion
            is_array($argumentResolvers) ? $argumentResolvers : array_values(iterator_to_array($argumentResolvers)),
```

to gurantee that the resulting array is a list.

DanielBadura (changes requested):
My suggested changes should fix the psalm errors. After that you would also need to update the code-style. `make cs` should do the trick.

src/Subscription/Subscriber/MetadataSubscriberAccessorRepository.php
gzumba: I put the array_values outside the whole expression, as we cannot trust the incoming array.

DanielBadura: We can trust the array here due to the type we defined. But it should be fine as is now.

DavidBadura (approved):

DanielBadura (approved):
Thanks for the PR @gzumba 👍

PR #612, opened 2024-07-09
infer aggregate id normalizer

DavidBadura:


DanielBadura (approved):

DanielBadura:
Regarding deptrac: would say this should be added to baseline. We already have there some issues regarding attributes.

PR #613, opened 2024-07-12
add write lock for event store

DavidBadura:
a fix for https://github.com/patchlevel/event-sourcing/issues/615

Implementation of write locking to ensure that the read part is always correct.

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -5,4 +5,6 @@ parameters:
 	level: max
 	paths:
 		- src
-	checkGenericClassInNonGenericObjectType: false
+	ignoreErrors:
+		-
+			identifier: missingType.generics
--- a/src/Store/DoctrineDbalStore.php
+++ b/src/Store/DoctrineDbalStore.php
@@ -61,6 +61,8 @@ final class DoctrineDbalStore implements Store, SubscriptionStore, DoctrineSchem
     /** @var array{table_name: string, aggregate_id_type: 'string'|'uuid', locking: bool, lock_id: int, lock_timeout: int} */
     private readonly array $config;
 
+    private bool $hasLock = false;
+
     /** @param array{table_name?: string, aggregate_id_type?: 'string'|'uuid', locking?: bool, lock_id?: int, lock_timeout?: int} $config */
     public function __construct(
         private readonly Connection $connection,
@@ -167,7 +169,7 @@ final class DoctrineDbalStore implements Store, SubscriptionStore, DoctrineSchem
         }
 
         $this->transactional(
-            function (Connection $connection) use ($messages): void {
+            function () use ($messages): void {
                 /** @var array<string, int> $achievedUntilPlayhead */
                 $achievedUntilPlayhead = [];
 
@@ -238,7 +240,7 @@ final class DoctrineDbalStore implements Store, SubscriptionStore, DoctrineSchem
                         continue;
                     }
 
-                    $this->executeSave($columns, $placeholders, $parameters, $types, $connection);
+                    $this->executeSave($columns, $placeholders, $parameters, $types, $this->connection);
 
                     $parameters = [];
                     $placeholders = [];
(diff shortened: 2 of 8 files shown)


src/Store/DoctrineDbalStore.php
```diff
@@ -423,4 +439,79 @@ private function executeSave(
...
+                    $this->config['lock_id'],
+                    $this->config['lock_timeout'],
+                ),
+            );
+
+            return;
+        }
+
+        if ($platform instanceof SQLitePlatform) {
+            return; // locking is not supported
```
DanielBadura: I think more accurate would be that locking is here not needed, right? Otherwise where is the distinction between sql lite and other db platforms which do not have a locking implementation for our users?

src/Store/DoctrineDbalStore.php
```diff
@@ -65,6 +73,9 @@ public function __construct(
         $this->config = array_merge([
             'table_name' => 'eventstore',
             'aggregate_id_type' => 'uuid',
+            'locking' => true,
+            'lock_id' => self::DEFAULT_LOCK_ID,
+            'lock_timeout' => -1,
```
DanielBadura, line 80: Maybe 
```php
[
  'lock' => [
      'enabled' => true,
      'id' => '',
      'timeout' => -1
   ]
]
```

phpstan.neon.dist
```diff
@@ -5,4 +5,6 @@ parameters:
 	level: max
 	paths:
 		- src
-	checkGenericClassInNonGenericObjectType: false
+	ignoreErrors:
+		-
+			identifier: missingType.generics
```
DanielBadura, line 10: cant we add theses errors to the baseline instead? This hides all these errors without the possibility then to fix them

DavidBadura: We don't want to fix these errors at all. And there are 75 errors, which would make the baseline quite large.

src/Store/DoctrineDbalStore.php
DavidBadura: I would like to leave the config here flat, anything else just makes it much more complicated.

phpstan.neon.dist
DanielBadura: I know checked it out and had a look on these errors. I know now what you mean. Ideally we would want to fix them, but it's not so easy and kinda tedious tbh.

src/Store/DoctrineDbalStore.php
DanielBadura: Im not sure if it makes it really complicated. I thought more about clustering it since the 3 option are tied together and makes it clearer from a user pov. I'm fine with both.

phpstan.neon.dist
DavidBadura: I didn't add this rule, I just fixed the deprecation. Can we discuss this separately whether it makes sense?

src/Store/DoctrineDbalStore.php
DavidBadura: This makes it more difficult because you have to deep merge defaults. And it is actually normal to make such configs flat, see `dbal connection` and `symfony form`.

phpstan.neon.dist
DanielBadura: Yeah - im fine with that config for now :)

src/Store/DoctrineDbalStore.php
DavidBadura: I have to look again at how it works with sqlite. I think it has a file lock by default, or you do it via file lock.

My idea was that DX is better, that you don't have to deal with locking when you switch between sqlite and other DBs for dev or testing. But I'm not sure about that.

DanielBadura: I think locking in SQLite is not needed due to the nature of how SQLite is working. What I meant here is that the comment is not quite correct. I would go a little more in detail that it is not needed and not `not supported`.

DanielBadura (approved):

PR #616, opened 2024-07-16
Improve Release Notifications

DanielBadura:
I rephrased the message a bit and added some hashtags for a possible better reach.


.github/workflows/notify-on-release.yml
```diff
@@ -4,6 +4,11 @@ on:
   release:
     types: [published]
 
+env:
+  NOTIFY_MESSAGE: |
+    We released ${{ github.event.release.tag_name }} of ${{ github.event.repository.name }}! See here for the changelog: ${{ github.event.release.html_url }}
+    #PHP #EventSourcing
```
DanielBadura:
```suggestion
    #PHP #DDD #EventSourcing
```

PR #620, opened 2024-07-24
[Experimental] Stream Store

### LINKED ISSUE #535: Open event store for non-aggregate events
Currently the event store can only be used with aggregates. This limits the use of the subscription system. A hybrid approach of d event sourcing and doctrine orm with domain events is also made more difficult.

Maybe we can generalize the store a bit so that we can also store non-aggregate events.

The following needs to be solved:

1) We don't have aggregate_name and aggregate_id. But we could probably solve it very easily by combining the fields in "stream", which is basically structured like this: "{aggregate_name}-{aggregate_id}". Then other systems could name their own stream such as: “doctrine” or “external”.

2) We also have to make playhead optional because it doesn't always exist. We set the column as "nullable". We have to check if we can do this for all DB platforms regarding the unqiue constraint.

DavidBadura:
This pull request introduces an event store that is decoupled from the aggregate root. This means that the columns `aggregate` and `aggregateId` have been merged into `stream`. Playhead is also optional in this store.

fix: https://github.com/patchlevel/event-sourcing/issues/535

### DIFF
--- a/docs/pages/store.md
+++ b/docs/pages/store.md
@@ -8,7 +8,6 @@ Each message contains an event and the associated headers.
     More information about the message can be found [here](message.md).
     
 The store is optimized to efficiently store and load events for aggregates.
-We currently only offer one [doctrine dbal](https://www.doctrine-project.org/projects/dbal.html) store.
 
 ## Create DBAL connection
 
@@ -29,8 +28,14 @@ $connection = DriverManager::getConnection(
     
 ## Configure Store
 
+We currently offer two stores, both based on the [doctrine dbal](https://www.doctrine-project.org/projects/dbal.html) library.
+The default store is the `DoctrineDbalStore` and the new experimental store is the `StreamDoctrineDbalStore`.
+
+### DoctrineDbalStore
+
+This is the current default store for event sourcing.
 You can create a store with the `DoctrineDbalStore` class.
-The store needs a dbal connection, an event serializer, an aggregate registry and some options.
+The store needs a dbal connection, an event serializer and has some optional parameters like options.
 
 ```php
 use Doctrine\DBAL\Connection;
@@ -41,21 +46,17 @@ use Patchlevel\EventSourcing\Store\DoctrineDbalStore;
 $store = new DoctrineDbalStore(
     $connection,
     DefaultEventSerializer::createFromPaths(['src/Event']),
-    null,
-    [/** options */],
 );
 ```
 Following options are available in `DoctrineDbalStore`:
 
-| Option            | Type             | Default    | Description                                  |
-|-------------------|------------------|------------|----------------------------------------------|
-| table_name        | string           | eventstore | The name of the table in the database        |
-| aggregate_id_type | "uuid"|"string" | uuid       | The type of the `aggregate_id` column        |
-| locking           | bool             | true       | If the store should use locking for writing  |
-| lock_id           | int              | 133742     | The id of the lock                           |
-| lock_timeout      | int              | -1         | The timeout of the lock. -1 means no timeout |
-
-## Schema
+| Option            | Type            | Default    | Description                                  |
+|-------------------|-----------------|------------|----------------------------------------------|
+| table_name        | string          | eventstore | The name of the table in the database        |
+| aggregate_id_type | "uuid"/"string" | uuid       | The type of the `aggregate_id` column        |
+| locking           | bool            | true       | If the store should use locking for writing  |
+| lock_id           | int             | 133742     | The id of the lock                           |
+| lock_timeout      | int             | -1         | The timeout of the lock. -1 means no timeout |
 
 The table structure of the `DoctrineDbalStore` looks like this:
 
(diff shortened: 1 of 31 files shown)


src/Message/Translator/RecalculatePlayheadTranslator.php
```diff
@@ -5,26 +5,52 @@
...
-        $playhead = $this->nextPlayhead($header->aggregateName, $header->aggregateId);
+        try {
+            $header = $message->header(AggregateHeader::class);
+        } catch (HeaderNotFound) {
+            try {
+                $header = $message->header(StreamHeader::class);
+            } catch (HeaderNotFound) {
+                return [$message];
+            }
+        }
```
DanielBadura, line 30: This is a change in the behaviour of this translator. Before, we crashed here if the header was not present and now we are returning a valid message. Is this fine?

src/Console/OutputStyle.php
```diff
@@ -97,4 +99,22 @@ public function throwable(Throwable $error): void
...
+    }
+
+    private function streamName(AggregateHeader|StreamHeader $header): string
+    {
+        if ($header instanceof AggregateHeader) {
+            return StreamNameTranslator::streamName($header->aggregateName, $header->aggregateId);
+        }
+
+        return $header->streamName;
+    }
```
DanielBadura: maybe this should go into the `StreamNameTranslator` itself?  Since this is done in another places in a different style

src/Message/Translator/RecalculatePlayheadTranslator.php
DavidBadura: I think it's OK. I find it strange when it crashes when no suitable header is found.

In this case, it's supposed to "fix" the playhead. But if there's no playhead, or no header with a playhead, then it has nothing to fix. Why should it stop here?

src/Console/OutputStyle.php
DavidBadura: Then we have an dependency `aggregate->message`.

Currently, the aggregate knows nothing about messages. (Except for the `AggregateHeader`, which will be removed in the long term)

In 4.0, `StreamNameTranslator` can also be removed again; it is mainly needed as a BC layer.
I want to solve this with metadata.

src/Message/Translator/RecalculatePlayheadTranslator.php
DanielBadura: I'm not sure anymore why we decided that tbh. But i think we had a reason. I agree with you, if nothing is there, then nothing to fix. I guess we implemented it that way, because a playhead was necessary. So, i think this is fine now. 👍

src/Console/OutputStyle.php
DanielBadura: You mean domain wise? I mean, the Translator could also be moved into `Store`, which is imho even a better place, no? 

I know that this translator acts as a BC layer, if the remove the "old" store in favor of the new `StreamStore`. I'm fine keeping it as is right now, but if we encounter a third place, i would vote for getting this more centralized.

docs/pages/store.md
```diff
@@ -41,21 +46,17 @@ use Patchlevel\EventSourcing\Store\DoctrineDbalStore;
...
-| lock_timeout      | int              | -1         | The timeout of the lock. -1 means no timeout |
-
-## Schema
+| Option            | Type            | Default    | Description                                  |
+|-------------------|-----------------|------------|----------------------------------------------|
+| table_name        | string          | eventstore | The name of the table in the database        |
+| aggregate_id_type | "uuid"/"string" | uuid       | The type of the `aggregate_id` column        |
+| locking           | bool            | true       | If the store should use locking for writing  |
+| lock_id           | int             | 133742     | The id of the lock                           |
+| lock_timeout      | int             | -1         | The timeout of the lock. -1 means no timeout |
```
DanielBadura, line 59: I created an issue to document more options for the bundle, see https://github.com/patchlevel/event-sourcing-bundle/issues/218

phpbench.json
```diff
@@ -43,7 +43,7 @@
                   "partition": "subject_name",
                   "cols":
                   {
-                    "time-diff": "percent_diff(partition['result_time_avg'][1], partition['result_time_avg'][0])"
+                    "time-diff": "percent_diff(coalesce(partition['result_time_avg']?[1], 0), coalesce(partition['result_time_avg']?[0], 0))"
```
DanielBadura, line 46: 🧠

DanielBadura (approved):

DanielBadura:
One last thing: Maybe we should add somewhere a note what to expect from a experimental feature? Like no BC compliance etc?

DavidBadura:
We should generally communicate how we handle our versions. Which versions are still supported, etc.

I opened an issue: https://github.com/patchlevel/event-sourcing/issues/622

src/Console/OutputStyle.php
DavidBadura: How the stream name is composed is the task of the Aggregate/Repository domain. The store should not know anything about aggregate. In the future, only the repository will create this string. So, this is not a BC layer for the store, but a BC layer for the repository/aggregate which store it can work with.

I don't know if the Aggregate namespace is the right place. Maybe repository? But in my opinion, definitely not store.

PR #632, opened 2024-09-19
add batch subscription

DavidBadura:
This PR allows batching of events into the subscriber. This can provide significant performance benefits depending on how the user implements it. A new interface is added called `BatchableSubscriber`, which gives you several methods to control exactly what should happen. The subscriber engine then takes care of the rest. 

Here is an example:

```php
use Doctrine\DBAL\Connection;
use Patchlevel\EventSourcing\Attribute\Projector;
use Patchlevel\EventSourcing\Subscription\Subscriber\BatchableSubscriber;

#[Projector('profile_1')]
final class MigrationSubscriber implements BatchableSubscriber
{
    public function __construct(
        private readonly Connection $connection,
    ) {
    }

    /** @var array<string, int> */
    private array $nameChanged = [];

    #[Subscribe(NameChanged::class)]
    public function handleNameChanged(NameChanged $event): void
    {
        $this->nameChanged[$event->userId] = $event->name;
    }

    public function beginBatch(): void
    {
        $this->nameChanged = [];
        $this->connection->beginTransaction();
    }

    public function commitBatch(): void
    {
        foreach ($this->nameChanged as $userId => $name) {
            $this->connection->executeStatement(
                'UPDATE user SET name = :name WHERE id = :id',
                ['name' => $name, 'id' => $userId],
            );
        }

        $this->connection->commit();
        $this->nameChanged = [];
    }

    public function rollbackBatch(): void
    {
        $this->connection->rollBack();
    }

    public function forceCommit(): bool
    {
        return count($this->nameChanged) > 1000;
    }
}
```


docs/pages/subscription.md
```diff
@@ -426,6 +426,96 @@ final class MigrationSubscriber
...
+        $this->connection->rollBack();
+    }
+
+    public function forceCommit(): bool
+    {
+        return count($this->nameChanged) > 1000;
+    }
+}
+```
+This interface provides you with all the options you need to process your data collectively.
```
DanielBadura, line 487:
```suggestion

This interface provides you with all the options you need to process your data collectively.
```

docs/pages/subscription.md
```diff
@@ -426,6 +426,96 @@ final class MigrationSubscriber
...
+The `commitBatch` method is called when batching was previously started, and one of the following conditions is met:
+Either the Subscription Engine reaches its limit, or the stream is finished.
+Alternatively, if the subscriber explicitly indicates using the `forceCommit` method that they want to process the data now.
+At this step, you must process all the data.
+
+The `rollbackBatch` method is called when an error occurs and the batching needs to be aborted.
+Here, you can respond to the error and potentially perform a database rollback.
+
+You can use the `forceCommit` method to indicate that you want to process the data now.
+This method is called after each handled event, and you can decide whether you want to process the data at that point.
```
DanielBadura: This and
```
Alternatively, if the subscriber explicitly indicates using the `forceCommit` method that they want to process the data now.
At this step, you must process all the data.
```
Is mostly the same. I think this should be combined down here.

src/Subscription/Engine/DefaultSubscriptionEngine.php
```diff
@@ -962,32 +988,109 @@ private function lowestSubscriptionPosition(array $subscriptions): int
...
+                $subscription->id(),
+                $e->getMessage(),
+                $e,
+            );
+        }
+
+        return null;
+    }
+
+    private function checkAndBeginBatch(Subscription $subscription): Error|null
```
DanielBadura, line 1062: Can't we place all the private batching methods into a internal class like you did with the `SubscriptionManager` and just init it in the constructor, not as a dep coming from outside?

src/Subscription/Subscriber/RealSubscriberAccessor.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Patchlevel\EventSourcing\Subscription\Subscriber;
+
+interface RealSubscriberAccessor
```
DanielBadura, line 7: This naming confuses me 🤔

src/Subscription/Engine/DefaultSubscriptionEngine.php
DavidBadura: This is not so easy to do, because otherwise I would have bidirectional dependencies between two classes. I want to avoid that.

src/Subscription/Subscriber/RealSubscriberAccessor.php
DavidBadura: Unfortunately, I can't think of anything better

DanielBadura (approved):

PR #637, opened 2024-10-06
filter events in subscription engine

DavidBadura:
This PR adds the option to filter the stream by events. This allows us to optimize the subscription engine so that we only load the events that the subscriptions need.

In other words, before the stream is loaded, all subscriptions that are currently being used are checked to see which events are needed. This list is then used to load the stream with a filter. As soon as a subscriber hears `*` (all), all events are always loaded.

This strategy is optional and can be tried out with the changed typehint for the store parameter in the subscription engine. The MessageLoader interface was introduced for this purpose.

```php
$engine = new DefaultSubscriptionEngine(
    new EventFilteredStoreMessageLoader(
        $store, 
        new AttributeEventMetadataFactory(), 
        $subscriberRepository,
    ),
    $subscriberRepository,
);
```

### DIFF
--- a/deptrac.yaml
+++ b/deptrac.yaml
@@ -154,6 +154,7 @@ deptrac:
       - Attribute
       - Clock
       - Message
+      - MetadataEvent
       - MetadataSubscriber
       - Repository
       - Schema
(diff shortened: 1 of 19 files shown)


DanielBadura (commented):
Maybe we should also add a `NOT IN` event filter?

DavidBadura:
For which use case :thinking:

DanielBadura:
Maybe for testing purposes, like checking what happens if an event will be dropped. It just came in my mind - nothing we need right away

DanielBadura (approved):

PR #638, opened 2024-10-15
add message header fallback if header can not be hydrated

DavidBadura:


DavidBadura:
I closed the PR because I came to the conclusion that it could cause more problems than it helps.

Headers are also often used for business rules. If these cannot be loaded and are automatically transformed into an `UnknownHeader`, then it can lead to major problems and data inconsistency.

In addition, the case that headers are not known is actually non-existent. It can only happen if the header class has been deleted or a bug has been introduced.

I can't think of any reason why you should delete a header. If you no longer need this header, then you don't use it anymore and you can filter it out with the pipeline in the future.

PR #641, opened 2024-10-16
dont teardown new subscriptions

DavidBadura:
Do not try to teardown new subscriptions because they have not been set up. This way there is nothing to teardown and the subscription can simply be deleted from the DB. Otherwise you always get an error message.


DanielBadura (approved):
CI fail is unrelated - should still wait until it is green.

PR #643, opened 2024-10-18
Add pipe & reducer

DavidBadura:
Add two new components: Pipe and Reducer

### Pipe

The `Pipe` is a construct that allows you to chain multiple translators.
This can be used to manipulate, filter or expand messages or events.
This can be used for anti-corruption layers, data migration, or to fix errors in the event stream.

```php
$messages = new Pipe(
    [$message1, $message2, /* ... */],
    new ExcludeEventTranslator(ProfileVisited::class)),
);

  // do something
}
```

### Reducer

The `Reducer` is a construct that allows you to reduce messages to a state.
This can be used to build temporal projections or to create a read model.

```php
$state = (new Reducer())
    ->initState(['name' => 'unknown'])
    ->match([
        ProfileCreated::class => static function (Message $message): array {
            return ['name' => $message->event()->name];
        },
        NameChanged::class => static function (Message $message): array {
            return ['name' => $message->event()->name];
        },
    ])
    ->reduce(
        new Pipe(
            $store->load(new Criteria(
                new AggregateIdCriterion($profileId->toString()),
                new AggregateNameCriterion('profile'),
            )),
            new UntilEventTranslator(new DateTimeImmutable()),
        ),
    );
```


src/Message/Pipe.php
```diff
@@ -0,0 +1,66 @@
...
+     * @return Generator<Message>
+     */
+    private function createGenerator(iterable $messages, Translator $translator): Generator
+    {
+        foreach ($messages as $message) {
+            $result = $translator($message);
+
+            foreach ($result as $m) {
+                yield $m;
+            }
```
DanielBadura:
```suggestion
            yield from $translator($message);
```

This should also be possible, right?

DanielBadura (approved):

Issue #230, opened 2022-04-24
Look again at naming to remove the "Default" prefix


DanielBadura:
We should try to get rid of the `Default*` naming scheme since it's not providing any useful information for the user. Instead we should think about to clearify what the class actually does, even if it is the default behaviour right now.

DanielBadura:
Not all default could be removed but [some were cleaned up](https://github.com/patchlevel/event-sourcing/pull/242). Still some classes with the prefix therefore not closing. Feel free to add suggestions!

PR #644, opened 2024-11-08
Add command bus, Drop PHP 8.1

DavidBadura:
```php
use Patchlevel\EventSourcing\Aggregate\BasicAggregateRoot;
use Patchlevel\EventSourcing\Attribute\Aggregate;
use Patchlevel\EventSourcing\Attribute\Apply;
use Patchlevel\EventSourcing\Attribute\Handle;
use Patchlevel\EventSourcing\Attribute\Id;

#[Aggregate('profile')]
final class Profile extends BasicAggregateRoot
{
    #[Id]
    private ProfileId $id;
    private string $name;

    #[Handle]
    public static function create(CreateProfile $command, NameValidator $nameValidator): self
    {
        $self = new self();
 
        if (!$nameValidator($command->name) {
            throw new InvalidArgument();
        }

        $self->recordThat(new ProfileCreated($command->id, $command->name));

        return $self;
    }

    #[Handle]
    public function changeName(ChangeProfileName $command, NameValidator $nameValidator): void
    {
        if (!$nameValidator($command->name) {
            throw new InvalidArgument();
        }

        $this->recordThat(new NameChanged($command->name));
    }

    #[Apply]
    protected function applyProfileCreated(ProfileCreated $event): void
    {
        $this->id = $event->profileId;
        $this->name = $event->name;
    }

    #[Apply]
    protected function applyNameChanged(NameChanged $event): void
    {
        $this->name = $event->name;
    }

    public function name(): string
    {
        return $this->name;
    }
}
```

```php
final class CreateProfile
{
    public function __construct(
        public readonly ProfileId $id,
        public readonly string $name,
    ) {
    }
}
```

```php
use Patchlevel\EventSourcing\Attribute\Id;

final class ChangeProfileName
{
    public function __construct(
        #[Id]
        public readonly ProfileId $id,
        public readonly string $name,
    ) {
    }
}
```
```php
use Patchlevel\EventSourcing\CommandBus\AggregateHandlerProvider;
use Patchlevel\EventSourcing\CommandBus\ServiceLocator;
use Patchlevel\EventSourcing\CommandBus\SyncCommandBus;
use Patchlevel\EventSourcing\Metadata\AggregateRoot\AggregateRootRegistry;
use Patchlevel\EventSourcing\Repository\RepositoryManager;

/**
 * @var AggregateRootRegistry $aggregateRootRegistry
 * @var RepositoryManager $repositoryManager
 */
$commandBus = new SyncCommandBus(
    new AggregateHandlerProvider(
        $aggregateRootRegistry,
        $repositoryManager,
        new ServiceLocator([
            'name_validator' => new NameValidator(),
        ]), // or other psr-11 compatible container
    ),
);

$commandBus->dispatch(new CreateProfile($profileId, 'John'));
$commandBus->dispatch(new ChangeProfileName($profileId, 'John Doe'));
```


src/CommandBus/Handler/DefaultParameterResolver.php
```diff
@@ -0,0 +1,61 @@
...
+                yield $command; // first parameter is always the command
+
+                continue;
+            }
+
+            if (!$this->container) {
+                throw ServiceNotResolvable::missingContainer();
+            }
+
+            yield $this->container->get(self::serviceName($method, $parameter));
```
DanielBadura: Maybe a try catch the exception from `container->get()` and re-throw the exception with `ServiceNotResolvable::missingService($method, $parameter, $exception)` ?

docs/pages/command_bus.md
```diff
@@ -0,0 +1,332 @@
...
+}
+```
+
+!!! note
+
+    To use Service Handler you need to register the handler in the `ServiceHandlerProvider`.
+
+!!! tip
+
+    A class can have multiple handler methods.
```
DanielBadura:
```suggestion
    A class can have multiple handle methods.
```

docs/pages/command_bus.md
```diff
@@ -0,0 +1,332 @@
...
+### Aggregate Handler
+
+Another way to handle commands is to use the aggregates themselves.
+To do this, you need to mark the method that handles the command with the `#[Handle]` attribute.
+
+!!! note
+
+    The aggregates themselves are of course not a service. 
+    The AggregateHandlerProvider uses the aggregates to create the handlers for you. 
+    You can find out more about this in the providers.
```
DanielBadura:
```suggestion
    You can find out more about this in the providers section.
```

DanielBadura: I think a link would also be great

docs/pages/command_bus.md
```diff
@@ -0,0 +1,332 @@
...
+ */
+$commandBus = new DefaultCommandBus($handlerProvider);
+
+$commandBus->dispatch(new CreateProfile($profileId, 'name'));
+$commandBus->dispatch(new ChangeProfileName($profileId, 'new name'));
+```
+
+!!! note
+
+    The `DefaultCommandBus` is a synchronous command bus.
```
DanielBadura: Maybe we could name it `SyncCommandBus` then?

DavidBadura: I don't know, I don't really like the name.

DavidBadura: I think I would stick with that name. We also have a `DefaultEventBus`.

DanielBadura: Yes i know, but also there the naming is not good IMHO. We also have an open issue about the naming problem https://github.com/patchlevel/event-sourcing/issues/230.

`SyncCommandBus` is better describing what the class is actually doing. You even noted it here in the docs so that the user is aware of it. If it would be named like this, the note wouldnt be necessary as the user already knows it.

This is not a hard blocker from my side, but i think we should not add more `Default*` classes if we can name it better.

DavidBadura: The note would stay the same, as the point is to emphasize that a command is always fully processed before another command is executed.

But if you want, I can rename it.

PR #645, opened 2024-11-09
Add docs for our bc promise and our versioning

### LINKED ISSUE #622: document which version policy we are using
* We use semver
* Which versions do we support?
* What does experimental mean
* What does internal mean

DanielBadura:
closes #622


DanielBadura:
Should we merge this in one of our 2.x branches?

DavidBadura (approved):

PR #656, opened 2024-12-13
[Experimental] split `StreamHeader` into 3 separate headers

DavidBadura:
For the experimental stream store, only the stream name is required. 

The requirement that 4 values ​​must be present, like the `AggregateHeader` has, dont exists in `StreamHeader`. 
That's why I have now split this header into 3 headers, so we don't have to deal with nullable values.

* StreamNameHeader
* PlayheadHeader
* RecordedOnHeader


src/Store/StreamDoctrineDbalStore.php
```diff
@@ -223,23 +226,34 @@ function () use ($messages): void {
...
+                        $parameters[] = $this->clock->now();
+                    }
+
                     $types[$offset + 4] = $dateTimeType;
 
                     $streamStart = $message->hasHeader(StreamStartHeader::class);
 
-                    if ($streamStart) {
-                        $achievedUntilPlayhead[$streamHeader->streamName] = $streamHeader->playhead;
+                    if ($streamStart && $playhead) {
```
DanielBadura:
```suggestion
                    if ($streamStart && $playhead  !== null) {
```

DanielBadura (approved):

PR #657, opened 2024-12-13
[Experimental] introduce event id & fix split stream logic in stream store

DavidBadura:
This PR needs https://github.com/patchlevel/event-sourcing/pull/656

Each event now gets a unique ID when saved in the StreamStore. You can either set this externally using EventIdHeader or it is generated automatically internally.

Using the event ID, I was also able to solve the problem with the split stream and missing playhead. I adapted the archiving logic so that I can simply archive all messages in a stream that happened before a certain event ID.

### DIFF
--- a/src/Store/StreamDoctrineDbalStore.php
+++ b/src/Store/StreamDoctrineDbalStore.php
@@ -292,17 +292,34 @@ final class StreamDoctrineDbalStore implements StreamStore, SubscriptionStore, D
                     $this->executeSave($columns, $placeholders, $parameters, $types, $this->connection);
                 }
 
+                $subselect = null;
+
                 foreach ($achievedUntilEventId as $stream => $eventId) {
+                    if ($subselect === null) {
+                        if ($this->connection->getDatabasePlatform() instanceof MySQLPlatform) {
+                            $subselect = sprintf(
+                                'SELECT t.id FROM (SELECT * FROM %s) AS t WHERE t.event_id = :event_id',
+                                $this->config['table_name'],
+                            );
+                        } else {
+                            $subselect = sprintf(
+                                'SELECT id FROM %s WHERE event_id = :event_id',
+                                $this->config['table_name'],
+                            );
+                        }
+                    }
+
                     $this->connection->executeStatement(
                         sprintf(
                             <<<'SQL'
-                            UPDATE %1$s
+                            UPDATE %s
                             SET archived = true
                             WHERE stream = :stream
-                            AND id < (SELECT id FROM %1$s WHERE event_id = :event_id)
+                            AND id < (%s)
                             AND archived = false
                             SQL,
                             $this->config['table_name'],
+                            $subselect,
                         ),
                         [
                             'stream' => $stream,
--- a/src/Store/StreamDoctrineDbalStore.php
+++ b/src/Store/StreamDoctrineDbalStore.php
@@ -29,11 +29,13 @@ use Patchlevel\EventSourcing\Store\Criteria\EventsCriterion;
 use Patchlevel\EventSourcing\Store\Criteria\FromIndexCriterion;
 use Patchlevel\EventSourcing\Store\Criteria\FromPlayheadCriterion;
 use Patchlevel\EventSourcing\Store\Criteria\StreamCriterion;
+use Patchlevel\EventSourcing\Store\Header\EventIdHeader;
 use Patchlevel\EventSourcing\Store\Header\PlayheadHeader;
 use Patchlevel\EventSourcing\Store\Header\RecordedOnHeader;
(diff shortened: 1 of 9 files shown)


src/Store/StreamDoctrineDbalStore.php
```diff
@@ -283,21 +292,38 @@ function () use ($messages): void {
...
                 }
 
-                foreach ($achievedUntilPlayhead as $stream => $playhead) {
+                $subselect = null;
+
+                foreach ($achievedUntilEventId as $stream => $eventId) {
+                    if ($subselect === null) {
+                        if ($this->connection->getDatabasePlatform() instanceof MySQLPlatform) {
+                            $subselect = sprintf(
+                                'SELECT t.id FROM (SELECT * FROM %s) AS t WHERE t.event_id = :event_id',
```
DanielBadura, line 301: Why is there an additional sub select? `(SELECT * FROM %s)`

src/Store/StreamDoctrineDbalStore.php
```diff
@@ -366,10 +392,13 @@ public function configureSchema(Schema $schema, Connection $connection): void
             ->setNotnull(true);
         $table->addColumn('playhead', Types::INTEGER)
             ->setNotnull(false);
-        $table->addColumn('event', Types::STRING)
+        $table->addColumn('event_id', Types::STRING)
+            ->setLength(255)
+            ->setNotnull(true);
+        $table->addColumn('event_name', Types::STRING)
             ->setLength(255)
             ->setNotnull(true);
```
DanielBadura, line 400: Did you test if the migration will be generated correctly? Renaming `event` to `event_name` and adding a new column `event_id`?

src/Store/StreamDoctrineDbalStore.php
```diff
@@ -283,21 +292,38 @@ function () use ($messages): void {
...
+                foreach ($achievedUntilEventId as $stream => $eventId) {
+                    if ($subselect === null) {
+                        if ($this->connection->getDatabasePlatform() instanceof MySQLPlatform) {
+                            $subselect = sprintf(
+                                'SELECT t.id FROM (SELECT * FROM %s) AS t WHERE t.event_id = :event_id',
+                                $this->config['table_name'],
+                            );
+                        } else {
+                            $subselect = sprintf(
+                                'SELECT id FROM %s WHERE event_id = :event_id',
```
DanielBadura, line 306: We already had the possibilty to make such a query with `stream` & `playhead` combined. So an extra `eventId` would not be necessary for this to work, right?

Nonetheless i think the eventId could be useful.

src/Store/StreamDoctrineDbalStore.php
DavidBadura: Because mysql... This is what mysql needs. MariaDB fixed it right after the fork xD

src/Store/StreamDoctrineDbalStore.php
DavidBadura: IMHO: correctly generated renaming in migration is not relevant because the store is defined as experimental.
You also have another problem: `event_id` is required, which also doesn't work with existing data. The user has to do this himself.

src/Store/StreamDoctrineDbalStore.php
DavidBadura: Yes, it is necessary. 
In the stream store, playhead is optional and can be null. This means you cannot search for this combination.

src/Store/StreamDoctrineDbalStore.php
DavidBadura: See: https://dev.mysql.com/doc/refman/8.4/en/subquery-restrictions.html
(second example: UPDATE query)

src/Store/StreamDoctrineDbalStore.php
DanielBadura: I know it's experimental, but i still think we should make this upgrade as smooth as possible if we are introducing BC's or updating the tables.

Regarding the ID, yes i know the user needs to add these themselfs. We still should guide the user how he could approach this, since a classical way like this would not work. 😄 

```php
$allMessages = $store->load();

forearch ($allMessages as $message) {
    $message->withHeader(new EventIdHeader(Uuid::uuid7()->toString()));
}

$store->save($allMessages);
```

src/Store/StreamDoctrineDbalStore.php
DanielBadura: True, forgot again that playhead is not required there

src/Store/StreamDoctrineDbalStore.php
DavidBadura: I will write something about it in the release notes

DanielBadura (approved):

PR #659, opened 2024-12-17
Lookup

DavidBadura:
This allows you to look at the current event stream in your subscription.
To look up information that is only relevant now. It also ensures that you cannot see into the future.

```php
#[Subscribe(AuctionClosed::class)]
public function onAuctionClosed(AuctionClosed $event, Lookup $lookup): void
{
    $messages = $lookup
      ->currentAggregate()
      ->events(
        Sold::class,
        AuctionPublished::class
      )->fetch();
    
    $state = (new Reducer())
        ->initState(['description' => null, 'price' => null])
        ->match([
            Sold::class => static function (Message $message): array {
                return ['price' => $message->event()->price];
            },
            AuctionPublished::class => static function (Message $message): array {
                return ['description' => $message->event()->description];
            },
        ])
        ->reduce($messages);
        
    $this->connection->insert('expired_auctions', $state);
}
```

The API of the Lookup class is not finished yet. I am thinking about how to build the API for it to filter by aggregateId, etc.


src/Serializer/DefaultEventSerializer.php
```diff
@@ -53,6 +53,11 @@ public function deserialize(SerializedEvent $data, array $options = []): object
         return $this->hydrator->hydrate($class, $payload);
     }
 
+    public function eventRegistry(): EventRegistry
```
DanielBadura: I'm not in favor of adding this method, just to have it easier in the tests. The `createFromPaths` is there to ease the creation and usage of this class by creating the object with reasonable default dependencies but this makes accessing the internal dependencies harder. Adding then methods to access these deps is not the solution imho, the solution should be to create this class manually then and not relying on `createFromPaths`.

The bare minimum should be marking this internal, as no user should rely on this.

src/Message/Reducer.php
```diff
@@ -93,16 +97,32 @@ public function finalize(Closure $closure): self
...
      *
      * @return OUT|STATE
      * @psalm-return (OUT is STATE ? STATE : OUT)
      */
-    public function reduce(iterable $messages): array
+    public function reduce(iterable|null $messages = null): array
     {
         $state = $this->initState;
 
+        if ($messages === null) {
```
DanielBadura: Why are we changing the API how to provide the messages? I did not see any usage here for the extra `Reducer::messages` method. This seems to just complicate things, no?


In this example, `messagesA` is unused, this could lead to confusion.
```php
$reducer = (new Reducer)->messages($messagesA);

$reducers->reduce($messagesB);
```

src/Subscription/Lookup/Lookup.php
```diff
@@ -0,0 +1,189 @@
...
+        $self->backwards = true;
+
+        return $self;
+    }
+
+    public function fetchAll(): Stream
+    {
+        return $this->store->load(
+            $this->criteria->add(
+                new ToIndexCriterion($this->currentMessage->header(IndexHeader::class)->index),
```
DanielBadura: this could already happen in the constructor, so then it never happens this will be missed, no?

src/Subscription/Lookup/Lookup.php
```diff
@@ -0,0 +1,189 @@
...
+    private bool $backwards = false;
+
+    private Criteria $criteria;
+
+    public function __construct(
+        private readonly Store $store,
+        private readonly Message $currentMessage,
+        private readonly EventRegistry|null $eventRegistry = null,
+    ) {
+        $this->criteria = new Criteria();
```
DanielBadura:
```suggestion
        $this->criteria = (new Criteria())->add(new ToIndexCriterion($this->currentMessage->header(IndexHeader::class)->index));
```

DanielBadura (changes requested):

src/Message/Reducer.php
DavidBadura: reverted

DanielBadura (approved):

PR #671, opened 2025-01-06
Use jsonb when using postgresql

DanielBadura:
We should use `jsonb` when the user is using postgresql. This give the user better query capabilities on the json fields in the store.

When this PR is released https://github.com/doctrine/dbal/pull/6693 the migration tooling should pick this change up. So the user does not need to update the table themselfes besides of the normal call to `bin/console do:mi:di` & `bin/console do:mi:mi`.

### DIFF
--- a/src/Subscription/ThrowableToErrorContextTransformer.php
+++ b/src/Subscription/ThrowableToErrorContextTransformer.php
@@ -70,6 +70,10 @@ final class ThrowableToErrorContextTransformer
      */
     private static function transformTrace(array $trace): array
     {
+        if (array_key_exists('class', $trace) && is_string($trace['class'])) {
+            $trace['class'] = str_replace("\x00", '', $trace['class']);
+        }
+
         if (!array_key_exists('args', $trace)) {
             return $trace;
         }
--- a/src/Subscription/ThrowableToErrorContextTransformer.php
+++ b/src/Subscription/ThrowableToErrorContextTransformer.php
@@ -21,6 +21,7 @@ use function is_object;
 use function is_resource;
 use function mb_strlen;
 use function mb_substr;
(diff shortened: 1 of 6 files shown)


src/Subscription/ThrowableToErrorContextTransformer.php
```diff
@@ -56,6 +56,10 @@ private static function transformThrowable(Throwable $error): array
      */
     private static function transformTrace(array $trace): array
     {
+        if (array_key_exists('class', $trace) && is_string($trace['class'])) {
+            $trace['class'] = str_replace("\x00", '', $trace['class']);
```
DanielBadura, line 75: Any better suggestions @DavidBadura ?

DavidBadura: Shouldn't that be its own PR?

DanielBadura: Right now, this is not doing any harm. When we adjust the table type to jsonb Postgres checks if the json is valid and doesn't like null bytes in there.

I don't think that this must be in a separate bugfix PR but I could extract it. The only thing is if the user already has null bytes in the table. Then he would need to remove them. Or do you have a case in mind?

DanielBadura:
New version for dbal was made which makes this change easier for our users.

Still need to evaluate if this is worth it due to slower write speed compared to the only limited gain for querying the eventstore.

DanielBadura:
I rebased this PR and would like to investigate into this again in the next days/weeks. 
Also, i think that the changes for the subscription table are probably not really needed and therefore the null byte removal is also not needed. Why? Because this is a real edge case to query the subscription table manually with advanced json features on the stack trace.

PR #687, opened 2025-02-06
Use reflection for accessing child aggregates, add test case which covers private child aggregates

### LINKED ISSUE #686: [experimental] Child Aggregates
Hello,

I would like first that I love your solution, and played with it since a week now !

I've played with basic features, but now I'm trying the experimental Child Aggregates feature, that match my future need.

I've encounter a potential design issue that I've discovered by chance (by testing).

Rather than provide a source code to demonstrate the problem you can see it yourself with your test code : 
https://github.com/patchlevel/event-sourcing/blob/3.8.x/tests/Integration/ChildAggregate/Profile.php

I used version 3.8.0 and test it under latest 3.8.1 too.

Now details of design error (for me) : 

Because the BasicAggregateRoot (`Profile` in test code) class is marked as final (that I understand and adheres), I don't see any logical reasons to put `protected` visibility on child aggregate (`PersonalInformation` instance) property.

So in my own structure I've built it with `private` visibility by default and found this error :  
`Error: Value of type null is not callable`

I've investigated and found that the recorder was not set in such use case.
And your test code it's reproductible  (when you define `private PersonalInformation $personalInformation;` in your Profile class)

```text
3) Patchlevel\EventSourcing\Tests\Integration\ChildAggregate\ChildAggregateIntegrationTest::testSnapshot
Error: Value of type null is not callable

/shared/backups/github/event-sourcing/src/Aggregate/ChildAggregateBehaviour.php:18
/shared/backups/github/event-sourcing/tests/Integration/ChildAggregate/PersonalInformation.php:31
/shared/backups/github/event-sourcing/tests/Integration/ChildAggregate/Profile.php:47
/shared/backups/github/event-sourcing/tests/Integration/ChildAggregate/ChildAggregateIntegrationTest.php:156
```

I'll let you explore this issue, but I think it came from following implementation 
https://github.com/patchlevel/event-sourcing/blob/3.8.x/src/Aggregate/AggregateRootAttributeBehaviour.php#L69-L71

That continue because property that support the child aggregate is not visible (not set) in this context, and the recorder is not set (that lead to error `Value of type null is not callable`).

DanielBadura:


src/Aggregate/AggregateRootAttributeBehaviour.php
```diff
@@ -49,9 +49,11 @@ protected function apply(object $event): void
         if (count($parts) === 2) {
             [$property, $method] = $parts;
 
-            /** @var ChildAggregate $child */
-            $child = $this->$property;
-            $child->$method($event);
+            $child = $this->getChildProperty($property);
```
DavidBadura: The name is confusing, it's not the property that is returned but the ChildAggregate. `getChildAggregate` (or `getChildAggregateByProperty`) would be more correct here, wouldn't it?

DavidBadura (approved):

PR #696, opened 2025-02-17
Replace psalm plugin "phinal" with phpstan extension "phpat"

DavidBadura:


DanielBadura (commented):
For the `final` check we don't need anymore a extra tool, since psalm is already checking it itself.

If we want to drop psalm then this might get relevant again. I would say we should add this as soon we are dropping psalm.

DavidBadura:
It's also about replacing deptrac.

DanielBadura:
Let's get this rolling. Replacing deptrac can be done another time :) 
I rebased this branch to latest minor.

DanielBadura (approved):

PR #697, opened 2025-02-17
Print the result into summary

DanielBadura:


DanielBadura:
Let's close for now as this is also not really visible.

PR #702, opened 2025-02-19
SubscribeMetadata dont hold a list of methods for events

### LINKED ISSUE #655: Disallow Subscriber listening on the same event multiple times
We enabled to have multiple `#[Subscribe(Event::class)]` in one Subscriber when we added the feature to listen to all events with `*` notation, [see here](https://github.com/patchlevel/event-sourcing/pull/498/files#diff-cfc5baa62f97bb1b59c3d1a0fd526c066e20bd84bca1cd8f27a137d541db6ffeL50-L66). This leads to the situation that the subscription engine cannot work 100% perfectly. What does that mean? If a Subscriber listens to one Event with multiple methods it can lead to unwanted actions. Let's take this Subscriber as an example:

```php
#[Projection('order')]
class OrderProjection 
{
   // ...
   
   #[Subscribe(OrderSubmitted::class)]
   public function sendConfirmationMail(OrderSubmitted $event): void 
   {
       // send mail
   }
   
   #[Subscribe(OrderSubmitted::class)]
   public function updateOrderInReadModel(OrderSubmitted $event): void 
   {
       // write to read model
   }
}
```

The described Subscribe may be defined as an antipattern here, as it is talking to 2 different systems here, but this is not the point of the Issue.

Now, if one of the 2 methods is failing, like e.g. the DB is not reachable, then the projection fails, which is normally no problem. But in this case, we dont know - or better the subcription engine does not know - which of these 2 methods failed. Since they are both listening on the same event the `position` is not helpful here. The result is, that both of these methods will be called again when retried. That could lead to either sending the same mail multiple times or corrupting the database with maybe false information if we are doing additive/mathy operations.

As I mentioned earlier: IMHO this situation would not occur if the software design would be better here in the application. So you could argue it is not a big issue then. But I can also not think of an valid usecase to allow this subscriber configuration. So I think we should not allow it at all.

Does it affect `*` notation? Yes, I think so, since we would have the same problem again if the have a method on the subscriber which listens on all events and some dedicated methods which are listening to some events. We would encounter the exact same problem again.

So my idea would be to disallow listening to the same event multiple times in a subscriber. As soon a subscriber listens to all events via `*` it cannot have other methods listening to any event.

DanielBadura:
As mentioned in https://github.com/patchlevel/event-sourcing/pull/701 this PR adjusts the API so that the metadata isnt holding a list of methods for an event anymore as this is not supported.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -439,7 +439,7 @@ parameters:
 			path: tests/Unit/Message/Translator/ReplaceEventTranslatorTest.php
 
 		-
-			message: '#^Method class@anonymous/tests/Unit/Metadata/Subscriber/AttributeSubscriberMetadataFactoryTest\.php\:212\:\:profileVisited\(\) has parameter \$message with no type specified\.$#'
+			message: '#^Method class@anonymous/tests/Unit/Metadata/Subscriber/AttributeSubscriberMetadataFactoryTest\.php\:211\:\:profileVisited\(\) has parameter \$message with no type specified\.$#'
 			identifier: missingType.parameter
 			count: 1
 			path: tests/Unit/Metadata/Subscriber/AttributeSubscriberMetadataFactoryTest.php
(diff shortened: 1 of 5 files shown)


DanielBadura:
Not sure why no BC-Break is here detected.

DavidBadura (approved):

PR #706, opened 2025-02-25
Add query bus

DanielBadura:


DavidBadura (changes requested):
Code looks good :+1: But the documentation is missing.

DavidBadura (approved):

Issue #712, opened 2025-03-14
Enhance WatchCommand (`event-sourcing:watch`) by adding usage of ConsoleLogger on worker creation


llaville:
Hello,

I've noticed that inside all CLI commands, two already implements usage of ConsoleLogger : 

- https://github.com/patchlevel/event-sourcing/blob/3.10.3/src/Console/Command/SubscriptionBootCommand.php#L95
- https://github.com/patchlevel/event-sourcing/blob/3.10.3/src/Console/Command/SubscriptionRunCommand.php#L96

But one is still missing :

- https://github.com/patchlevel/event-sourcing/blob/3.10.3/src/Console/Command/WatchCommand.php#L95

Perharps it could be helpfull to have such same behaviour

A patch will look like this one 

```
diff --git a/src/Console/Command/WatchCommand.php b/src/Console/Command/WatchCommand.php
index 91fc67af..6cbcf87a 100644
--- a/src/Console/Command/WatchCommand.php
+++ b/src/Console/Command/WatchCommand.php
@@ -17,6 +17,7 @@ use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Logger\ConsoleLogger;
 use Symfony\Component\Console\Output\OutputInterface;

 #[AsCommand(
@@ -92,6 +93,8 @@ final class WatchCommand extends Command

         $criteria = $criteriaBuilder->build();

+        $logger = new ConsoleLogger($output);
+
         $worker = DefaultWorker::create(
             function () use ($console, &$index, $criteria, $sleep): void {
                 $stream = $this->store->load(
@@ -113,6 +116,8 @@ final class WatchCommand extends Command

                 $this->store->wait($sleep);
             },
+            [],
+            $logger
         );

         $supportSubscription = $this->store instanceof SubscriptionStore && $this->store->supportSubscription();		 
```

DanielBadura:
Yes, could be a good little improvement. Maybe you could open a PR for this?

llaville:
> Yes, could be a good little improvement. Maybe you could open a PR for this?

Ok to propose a PR. Will come under next 24 hours ...

llaville:
I'll push the PR even, if I got error (not related to my changes) , see follows, `make dev` results command : 

```
vendor/bin/psalm

[...]

PR #717, opened 2025-03-21
Add ConsoleLogger on Watch Command

llaville:
Hello,

This PR fixe #712 

- I've introduced a basic unit test for watch command that did not exists yet !

- I've also used the InMemoryStore in unit test because I cannot prophetysed the Store

With this code
```php
$store = $this->prophesize(Store::class);
```

I got this error 
```
TypeError: Double\Store\P1::load(): Return value must be of type Patchlevel\EventSourcing\Store\Stream, null returned
```

### DIFF
--- a/src/Console/Command/WatchCommand.php
+++ b/src/Console/Command/WatchCommand.php
@@ -17,6 +17,8 @@ use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Logger\ConsoleLogger;
+use Symfony\Component\Console\Output\ConsoleOutputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
 #[AsCommand(
@@ -36,6 +38,24 @@ final class WatchCommand extends Command
     protected function configure(): void
     {
         $this
+            ->addOption(
+                'run-limit',
+                null,
+                InputOption::VALUE_REQUIRED,
+                'The maximum number of runs this command should execute',
+            )
+            ->addOption(
+                'memory-limit',
+                null,
+                InputOption::VALUE_REQUIRED,
+                'How much memory consumption should the worker be terminated (e.g. 250MB)',
+            )
+            ->addOption(
+                'time-limit',
+                null,
+                InputOption::VALUE_REQUIRED,
+                'What is the maximum time the worker can run in seconds',
+            )
             ->addOption(
                 'sleep',
                 null,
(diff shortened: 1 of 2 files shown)


src/Console/Command/WatchCommand.php
```diff
@@ -67,6 +87,9 @@ protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $console = new OutputStyle($input, $output);
 
+        $runLimit = InputHelper::nullablePositiveInt($input->getOption('run-limit'));
+        $memoryLimit = InputHelper::nullableString($input->getOption('memory-limit'));
+        $timeLimit = InputHelper::nullablePositiveInt($input->getOption('time-limit'));
```
DanielBadura, line 92: I'm not sure if these options are really useful for the watcher, as the current usecase is only in dev. There i dont think you really want to stop the worker e.g. after 10 minutes or so. On the other hand it is not really harmful either and the maintenance cost should be low, wdyt @DavidBadura

src/Console/Command/WatchCommand.php
```diff
@@ -92,6 +115,9 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 
         $criteria = $criteriaBuilder->build();
 
+        $errOutput = $output instanceof ConsoleOutputInterface ? $output->getErrorOutput() : $output;
```
DanielBadura, line 118: Why are we passing only the error output here?

DanielBadura (commented):
The problem with the mocking of the `Store` was probably that you didnt mock the `load` method and that's why it crashed. But testing with the in memory store is also fine.

src/Console/Command/WatchCommand.php
llaville: Like with Symfony, I'd prefer to have logs wrote on STDERR rather than STDOUT (fallback to)

src/Console/Command/WatchCommand.php
DavidBadura: I also think it doesn't make sense to set any limits in general, since it's just a dev command. It doesn't hurt to maintain it, but it doesn't add any value either.

I've only found one case where you initially think it makes sense, but then you quickly realize that it only provides a false sense of security:

That case is when the limit is exceeded and the command is shut down cleanly and restarted by something else (like Docker or Superviser). However, there's the problem that events are thrown between shutdown and restart, which are then not displayed in the console. In other words, you "lose" events when displaying them, which I find very confusing. So I wouldn't just stop / restart the command secretly, but would indicate that something went wrong.

I would tend not to add the limits, but I'm fine with adding them.

src/Console/Command/WatchCommand.php
DavidBadura: I also checked, this is also done in Symfony Messenger. Everything is written to std error there. I honestly don't know when I would write something to std error or std out, so I had everything in std out. But I think it makes sense to put logs in std error, at least for logs. 👍

DavidBadura: @DanielBadura I think we should even do the same with the other commands.

src/Console/Command/WatchCommand.php
DavidBadura: @llaville Maybe do you have a use case that I have overlooked?

llaville: @DavidBadura This is for test case that doesn't exists yet on WatchCommand I've introduced : see https://github.com/llaville/event-sourcing/commit/143c2e659e715b271108bd1beaa01eba25dfc00d#diff-d2abab7733f01348ecec5ddc13ecbe1cb84128e85be13db41fcea44cadfca9ecR42 (line 42)

DavidBadura (approved):

DanielBadura (approved):
Thank you!

PR #720, opened 2025-03-23
Typo fix about Aggregate Root Registry

llaville:
Hello, 

Another fix on current documentation (3.10.4), about usage of Aggregate Root Registry


DanielBadura:
Thanks!

Issue #719, opened 2025-03-22
Enhance DebugCommand (`event-sourcing:debug`) by displaying subscribers if available


llaville:
Hello,

While I'm pretty happy of my current implementation of your library (current version 3.10.4), sometimes I would like to display all events, aggregates and subscriptions on my project with only one command.

With version 3.10.4, I should run at least two commands with [CLI](https://event-sourcing.patchlevel.io/latest/cli/) to do it : 

- `event-sourcing:debug` (not yet documented) that will display only events and aggregates defined
- `event-sourcing:subscription:status` to display list of subscriptions if available 

This last one suffer that the (default Doctrine) implementation does not checks if table (default `subscriptions`) exists (may raise a `Doctrine\DBAL\Exception\TableNotFoundException` by https://github.com/patchlevel/event-sourcing/blob/3.11.x/src/Subscription/Engine/DefaultSubscriptionEngine.php#L963)

i.e:
<details>
<summary>Exception Report</summary>

```text
devilbox@php-8.2.28 in /shared/backups/bartlett/php-compatinfo-db $ bin/es.php event-sourcing:subscription:status -v

In ExceptionConverter.php line 52:

  [Doctrine\DBAL\Exception\TableNotFoundException (1)]
  An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1 no such table: subscriptions


Exception trace:
  at /shared/backups/bartlett/php-compatinfo-db/vendor/doctrine/dbal/src/Driver/API/SQLite/ExceptionConverter.php:52
 Doctrine\DBAL\Driver\API\SQLite\ExceptionConverter->convert() at /shared/backups/bartlett/php-compatinfo-db/vendor/doctrine/dbal/src/Connection.php:1418
 Doctrine\DBAL\Connection->handleDriverException() at /shared/backups/bartlett/php-compatinfo-db/vendor/doctrine/dbal/src/Connection.php:1360
 Doctrine\DBAL\Connection->convertExceptionDuringQuery() at /shared/backups/bartlett/php-compatinfo-db/vendor/doctrine/dbal/src/Connection.php:787
 Doctrine\DBAL\Connection->executeQuery() at /shared/backups/bartlett/php-compatinfo-db/vendor/doctrine/dbal/src/Query/QueryBuilder.php:305
[...]

PR #721, opened 2025-03-24
Add subscriber information to `event-sourcing:debug`

llaville:
Hello,

Here is the PR to improve `debug:event-sourcing` command, following principles presented into #719

### DIFF
--- a/src/Console/Command/DebugCommand.php
+++ b/src/Console/Command/DebugCommand.php
@@ -7,18 +7,22 @@ namespace Patchlevel\EventSourcing\Console\Command;
 use Patchlevel\EventSourcing\Console\OutputStyle;
 use Patchlevel\EventSourcing\Metadata\AggregateRoot\AggregateRootRegistry;
 use Patchlevel\EventSourcing\Metadata\Event\EventRegistry;
+use Patchlevel\EventSourcing\Subscription\Subscriber\MetadataSubscriberAccessor;
+use Patchlevel\EventSourcing\Subscription\Subscriber\SubscriberAccessorRepository;
 use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
+use function array_intersect_key;
 use function array_keys;
 use function array_map;
+use function array_shift;
 use function array_values;
 
 #[AsCommand(
     'event-sourcing:debug',
-    'show event sourcing debug information',
+    'Show event sourcing information (aggregates, events, subscribers)',
     ['debug:event-sourcing'],
 )]
 final class DebugCommand extends Command
@@ -26,6 +30,7 @@ final class DebugCommand extends Command
     public function __construct(
         private readonly AggregateRootRegistry $aggregateRootRegistry,
         private readonly EventRegistry $eventRegistry,
+        private readonly SubscriberAccessorRepository|null $subscriberAccessorRepository = null,
     ) {
         parent::__construct();
     }
@@ -36,8 +41,9 @@ final class DebugCommand extends Command
 
         $this->showAggregates($console);
         $this->showEvents($console);
+        $this->showSubscribers($console);
 
-        return 0;
+        return Command::SUCCESS;
     }
 
     private function showAggregates(OutputStyle $console): void
@@ -61,6 +67,45 @@ final class DebugCommand extends Command
         $console->table(
             ['name', 'class'],
             array_map(null, array_keys($events), array_values($events)),
+        );
+    }
+
+    private function showSubscribers(OutputStyle $console): void
+    {
+        $eventNames = $this->eventRegistry->eventNames();
+
+        $subscribers = [];
+
+        /** @var MetadataSubscriberAccessor $subscriberAccessor */
+        foreach ($this->subscriberAccessorRepository?->all() ?? [] as $subscriberAccessor) {
+            $metadata = $subscriberAccessor->metadata();
+
+            $eventsHandled = array_values(array_intersect_key($eventNames, $metadata->subscribeMethods));
+
+            $subscribers[] = [
+                $metadata->id,
+                $metadata->group,
+                $metadata->runMode->value,
+                // first event name handled
+                array_shift($eventsHandled),
+            ];
+
+            // display more event names that the subscriber can handle (if any)
+            foreach ($eventsHandled as $eventName) {
+                $subscribers[] = ['', '', '', $eventName];
+            }
+        }
+
+        $console->title('Subscribers');
+
+        $console->table(
+            [
+                'id',
+                'group',
+                'run mode',
+                'events handled',
+            ],
+            $subscribers,
         );
     }
 }
--- a/src/Console/Command/DebugCommand.php
+++ b/src/Console/Command/DebugCommand.php
@@ -71,8 +71,6 @@ final class DebugCommand extends Command
 
     private function showSubscribers(OutputStyle $console): void
     {
-        $eventNames = $this->eventRegistry->eventNames();
-
         $subscribers = [];
 
         /** @var MetadataSubscriberAccessor $subscriberAccessor */
@@ -85,11 +83,11 @@ final class DebugCommand extends Command
                 $metadata->id,
                 $metadata->group,
                 $metadata->runMode->value,
-                // first event name handled
+                // first event class name handled
                 array_shift($eventsHandled),
             ];
 
-            // display more event names that the subscriber can handle (if any)
+            // display more event class names that the subscriber can handle (if any)
             foreach ($eventsHandled as $eventName) {
                 $subscribers[] = ['', '', '', $eventName];
             }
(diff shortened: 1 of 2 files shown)


DavidBadura:
Hey, Thanks for the PR! 

I added a note to the ticket at the same time, sorry for the delay. Can you take a look? Thanks!

src/Console/Command/DebugCommand.php
```diff
@@ -4,28 +4,33 @@
...
-    'show event sourcing debug information',
+    'Show event sourcing information (aggregates, events, subscriptions)',
     ['debug:event-sourcing'],
 )]
 final class DebugCommand extends Command
 {
     public function __construct(
         private readonly AggregateRootRegistry $aggregateRootRegistry,
         private readonly EventRegistry $eventRegistry,
+        private readonly SubscriptionEngine $engine,
```
DavidBadura: This must be optional (nullable) because of our BC promise.

llaville: Fixed by new implementation that show subscribers instead of subscriptions (commit 084fbbdffa8c07c83cab7649d1c63df48cb209c4)

tests/Unit/Console/Command/DebugCommandTest.php
```diff
@@ -37,4 +46,37 @@ public function testSuccessful(): void
...
+        $output = new BufferedOutput();
+
+        $exitCode = $command->run($input, $output);
+
+        self::assertSame(0, $exitCode);
+
+        $content = $output->fetch();
+
+        self::assertStringContainsString('default', $content);
+        self::assertStringContainsString('projector', $content);
```
DanielBadura: Could you add asserts the other rows of the table as well? This would probably fix most if not all of the infection issues on the diff.

src/Console/Command/DebugCommand.php
```diff
@@ -63,4 +69,43 @@ private function showEvents(OutputStyle $console): void
...
+    {
+        $eventNames = $this->eventRegistry->eventNames();
+
+        $subscribers = [];
+
+        /** @var MetadataSubscriberAccessor $subscriberAccessor */
+        foreach ($this->subscriberAccessorRepository?->all() ?? [] as $subscriberAccessor) {
+            $metadata = $subscriberAccessor->metadata();
+
+            $eventsHandled = array_values(array_intersect_key($eventNames, $metadata->subscribeMethods));
```
DanielBadura: Here we are filtering the configured events from the subscriber by the registered events in the registry, right?

Not sure if we should do this here, as the subscriber defines that it is handling these events, even if they are not present in the registry. IMHO this would just hide information from the user. It may be a misconfiguration or intended but we should show it. WDYT @DavidBadura ?

We could think about listing these events differently, highlighting it with a warning color (yellow) if possible or in a dedicated column which indicates that these events are not registered.

DanielBadura (changes requested):

tests/Unit/Console/Command/DebugCommandTest.php
llaville: I've tried to improve it as much as possible (commit https://github.com/patchlevel/event-sourcing/pull/721/commits/0589e5615d904e2b92f3944f83eb3aa55cd4c8f1), but even with latest version of Infection 0.29.12 / PHPUnit 11.5.15, I got only (after applying the patch presented into #722)

```text
php -d memory_limit=312M vendor/bin/roave-infection-static-analysis-plugin --threads=max --git-diff-lines --git-diff-base=origin/HEAD --ignore-msi-with-no-mutations --only-covered --min-msi=80 --min-covered-msi=95

    ____      ____          __  _
   /  _/___  / __/__  _____/ /_(_)___  ____
   / // __ \/ /_/ _ \/ ___/ __/ / __ \/ __ \
 _/ // / / / __/  __/ /__/ /_/ / /_/ / / / /
/___/_/ /_/_/  \___/\___/\__/_/\____/_/ /_/

#StandWithUkraine

Infection - PHP Mutation Testing Framework version 0.29.12


Running initial test suite...

PHPUnit version: 11.5.15

  677 [============================] 19 secs

Generate mutants...

Processing source code files: 1/1
.: killed, M: escaped, U: uncovered, E: fatal error, X: syntax error, T: timed out, S: skipped, I: ignored

..M......M.                                          (11 / 11)

11 mutations were generated:
       9 mutants were killed
       0 mutants were configured to be ignored
       0 mutants were not covered by tests
       2 covered mutants were not detected
       0 errors were encountered
       0 syntax errors were encountered
       0 time outs were encountered
       0 mutants required more time than configured

Metrics:
         Mutation Score Indicator (MSI): 81%
         Mutation Code Coverage: 100%
         Covered Code MSI: 81%

Generated Reports:
         - /shared/backups/forks/event-sourcing/infection.log
         - /shared/backups/forks/event-sourcing/infection.html

Please note that some mutants will inevitably be harmless (i.e. false positives).
[warning] Dashboard report has not been sent: The current process is not executed in a CI build

Time: 20s. Memory: 0.07GB. Threads: 11

 [ERROR] The minimum required Covered Code MSI percentage should be 95%, but actual is 81.82%. Improve your tests!

```

src/Console/Command/DebugCommand.php
DavidBadura: I also think that we should display what the user has specified, regardless of whether it's in the registry or not. I also think we shouldn't use highlighting here. 

If we want to include such checks, we should do it in our phpstan/psalm plugin.

tests/Unit/Console/Command/DebugCommandTest.php
```diff
@@ -37,4 +46,42 @@ public function testSuccessful(): void
...
+
+        $content = $output->fetch();
+
+        self::assertStringContainsString('Subscribers', $content);
+        self::assertStringContainsString('events handled', $content);
+        self::assertStringContainsString('profile.test', $content);
+        self::assertStringContainsString('default', $content);
+        self::assertStringContainsString('profile.projection', $content);
+        self::assertStringContainsString('projector', $content);
+        self::assertStringContainsString('profile.created', $content);
```
DanielBadura:
```suggestion
        self::assertStringContainsString('Subscribers', $content);
        self::assertStringContainsString('id', $content);
        self::assertStringContainsString('group', $content);
        self::assertStringContainsString('run mode', $content);
        self::assertStringContainsString('events handled', $content);
        
        self::assertStringContainsString('profile.test', $content);
        self::assertStringContainsString('default', $content);
        self::assertStringContainsString('from_beginning', $content);
        
        self::assertStringContainsString('profile.projection', $content);
        self::assertStringContainsString('projector', $content);
        self::assertStringContainsString('from_beginning', $content);
        self::assertStringContainsString('profile.created', $content);
```

This should remove one of the issue reported by infection, the mutation was removing the `id` column in the table.

src/Console/Command/DebugCommand.php
```diff
@@ -63,4 +69,43 @@ private function showEvents(OutputStyle $console): void
...
+    {
+        $eventNames = $this->eventRegistry->eventNames();
+
+        $subscribers = [];
+
+        /** @var MetadataSubscriberAccessor $subscriberAccessor */
+        foreach ($this->subscriberAccessorRepository?->all() ?? [] as $subscriberAccessor) {
+            $metadata = $subscriberAccessor->metadata();
+
+            $eventsHandled = array_intersect_key($eventNames, $metadata->subscribeMethods);
```
DanielBadura: Let's remove the filtering of the events, as this command shows only configuration.

BTW: infection is also reporting this, since the tests dont cover it. You would need to add in the test a subscriber which configures an event which is not present in the registry.

DanielBadura (changes requested):
With these 2 comments, the CI should be green and this patch good to go.

src/Console/Command/DebugCommand.php
llaville: If I remove the filtering of events by : 
```
        foreach ($this->subscriberAccessorRepository?->all() ?? [] as $subscriberAccessor) {
            $metadata = $subscriberAccessor->metadata();

            $eventsHandled = \array_keys($metadata->subscribeMethods);

```

It will works for Infection without need to add a new unit test case, but the results is different 
In previous attempt, it display Event Name, and now it display Event Classname

I.e (in my project context), it gave something like 

```
Subscribers
===========

 -------------- ----------- ---------------- --------------------------------------------------------
  id             group       run mode         events handled
 -------------- ----------- ---------------- --------------------------------------------------------
  distribution   projector   from_beginning   Bartlett\CompatInfoDb\Domain\Event\DistributionCreated
                                              Bartlett\CompatInfoDb\Domain\Event\ExtensionCheckedIn
                                              Bartlett\CompatInfoDb\Domain\Event\ExtensionCheckedOut
 -------------- ----------- ---------------- --------------------------------------------------------
```
It's ok, even if Event Name has my preference !
WDYT about this last one ?

llaville: See new commits https://github.com/patchlevel/event-sourcing/pull/721/commits/7d55f75d61611945756060e6622b4d75b6d1dec4 and https://github.com/patchlevel/event-sourcing/pull/721/commits/a018d13c8bc18675840a29e3463030a6e50d6ad5 for changes

tests/Unit/Console/Command/DebugCommandTest.php
llaville: Resolved by commit https://github.com/patchlevel/event-sourcing/pull/721/commits/7d55f75d61611945756060e6622b4d75b6d1dec4

src/Console/Command/DebugCommand.php
DanielBadura: Yes this is fine.

DanielBadura (approved):
Looks great now, thank you!

DavidBadura (approved):

PR #751, opened 2025-06-17
add instant retry command bus

DavidBadura:
Replace the `RetryOutdatedAggregateCommandBus` with a more flexible version. The `InstantRetryCommandBus` allows you to define default values ​​and now also allows you to specify a list of exceptions that trigger this functionality, not just `OutdatedAggregate`.

### DIFF
--- a/src/Attribute/InstantRetry.php
+++ b/src/Attribute/InstantRetry.php
@@ -1,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Patchlevel\EventSourcing\Attribute;
+
+use Attribute;
+use Throwable;
+
+#[Attribute(Attribute::TARGET_CLASS)]
+final class InstantRetry
+{
+    /**
+     * @param positive-int|null                  $maxRetries
+     * @param list<class-string<Throwable>>|null $exceptions
+     */
+    public function __construct(
+        public readonly int|null $maxRetries = null,
+        public readonly array|null $exceptions = null,
+    ) {
+    }
+}
--- a/src/CommandBus/InstantRetryCommandBus.php
+++ b/src/CommandBus/InstantRetryCommandBus.php
@@ -1,0 +1,65 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Patchlevel\EventSourcing\CommandBus;
+
+use Patchlevel\EventSourcing\Attribute\InstantRetry;
+use Patchlevel\EventSourcing\Repository\AggregateOutdated;
+use ReflectionClass;
+use Throwable;
+
+use function in_array;
+
+final class InstantRetryCommandBus implements CommandBus
+{
+    /**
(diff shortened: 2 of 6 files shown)


src/Attribute/InstantRetry.php
```diff
@@ -0,0 +1,19 @@
...
+
+use Attribute;
+use Throwable;
+
+#[Attribute(Attribute::TARGET_CLASS)]
+final class InstantRetry
+{
+    /** @param list<class-string<Throwable>>|null $exceptions */
+    public function __construct(
+        public readonly int|null $maxRetries = null,
```
DanielBadura, line 18: Negative integers should not be allowed, could you add this type info?

src/CommandBus/InstantRetryCommandBus.php
```diff
@@ -0,0 +1,62 @@
...
+use Throwable;
+
+use function in_array;
+
+final class InstantRetryCommandBus implements CommandBus
+{
+    /** @param list<class-string<Throwable>> $defaultExceptions */
+    public function __construct(
+        private readonly CommandBus $commandBus,
+        private readonly int $defaultMaxRetries = 3,
```
DanielBadura, line 22: Here also: no negatives allowed

src/CommandBus/InstantRetryCommandBus.php
```diff
@@ -0,0 +1,62 @@
...
+            $configuration = $this->configuration($command);
+
+            if ($configuration === null) {
+                throw $exception;
+            }
+
+            $exceptions = $configuration->exceptions ?? $this->defaultExceptions;
+            $maxRetries = $configuration->maxRetries ?? $this->defaultMaxRetries;
+
+            if ($retry >= $maxRetries || !in_array($exception::class, $exceptions, true)) {
```
DanielBadura, line 46: If we allow negatives, this would result in a infinite loop, so let's dont allow them on type level

tests/Unit/CommandBus/InstantRetryCommandBusTest.php
```diff
@@ -0,0 +1,170 @@
...
+        $retryCommandBus = new InstantRetryCommandBus($innerCommandBus);
+
+        $retryCommandBus->dispatch($command);
+
+        $this->assertTrue(true); // If no exception is thrown, the test passes
+    }
+
+    public function testDispatchThrowsAfterMaxRetries(): void
+    {
+        $command = new #[InstantRetry(maxRetries: 2)]
```
DanielBadura, line 96: Could you duplicate this test without setting a max retry via attribute? This would then test the default retry value of the command bus which is currently untested. This should then resolve the report from infection

DanielBadura (changes requested):

DanielBadura (approved):
Nice I really like this approach!

PR #753, opened 2025-07-02
cleanup subscription engine constructor

DavidBadura:
remove deprecated constructor parameters in subscription engine


docs/pages/UPGRADE-4.0.md
```diff
@@ -0,0 +1,7 @@
+# Upgrade 4.0
+
+## Subscription
+
+* The constructor of the `SubscriptionEngine` class has been changed
+* * Instead of passing a `Store` instance, you now need to pass a `MessageLoader` instance.
+* * Instead of passing a `RetryStrategy` instance, you now need to pass a `RetryStrategyRepository` instance.
```
DanielBadura: I think we should be a little more presice here about how to upgrade besides using the other interface(s).

Like mentioning that there is a `StoreMessageLoader` which takes a `Store` as an argument e.g.

src/Subscription/RetryStrategy/RetryStrategyRepository.php
```diff
@@ -28,4 +28,22 @@ public function getDefaultRetryStrategy(): RetryStrategy
...
+    }
+
+    public static function noRetry(): self
+    {
+        return new self(
+            [
+                self::DEFAULT_STRATEGY_NAME => new NoRetryStrategy(),
+            ],
+            self::DEFAULT_STRATEGY_NAME,
+        );
```
DanielBadura: Two different indendation styles :D

DanielBadura (approved):

PR #755, opened 2025-07-03
remove experimental child aggregates

DavidBadura:
We removed our experimental feature of child aggregates. 
This was our first attempt to split aggregates into smaller parts,
but we found a better way to do this with the `Micro Aggregate` feature.


src/Metadata/AggregateRoot/AttributeAggregateRootMetadataFactory.php
```diff
@@ -286,7 +239,8 @@ private function getEventClassesByPropertyTypes(ReflectionMethod $method): array
 
         if ($propertyType instanceof ReflectionUnionType) {
             return array_map(
-                static function (ReflectionNamedType|ReflectionIntersectionType $reflectionType) use ($methodName): string {
+                static function (ReflectionNamedType|ReflectionIntersectionType $reflectionType) use ($methodName,
+                ): string {
```
DanielBadura: weird indentation

docs/pages/aggregate.md
```diff
@@ -653,7 +653,7 @@ Or for test purposes the `FrozenClock`, which always returns the same time.
 
 In some cases, it makes sense to split an aggregate into several smaller aggregates.
 This can be the case if the aggregate becomes too large or if the aggregate is used in different contexts.
-We currently support two patterns for this: Micro Aggregates and Child Aggregates (experimental).
+We currently support only one pattern for this: Micro Aggregates.
```
DanielBadura: I would rephrase this, as it contextually now not much sense to talk about only one pattern.

```suggestion
For these cases you can use Micro Aggregates.
```

DanielBadura (changes requested):

DanielBadura (approved):

PR #763, opened 2025-07-28
DCB

DavidBadura:
Implement Dynamic Consistency Boundary according to https://dcb.events/.

The following examples were partially implemented:

* [Course subscription](https://dcb.events/examples/course-subscriptions/)
* [Invoice number](https://dcb.events/examples/invoice-number/)

There are still some questions open:

* [x] How do we handle the different streams? It should be possible to use different streams for different bounded contexts. Currently, this is not reflected in the API. -> implemented
* [x] The `EventTagExtractor` must handle value objects. The current implementation requires PHP's Stringable interface, which requires a `__toString()` method. Since this can lead to problems, we need an alternative. -> use Identifier component
* [x] Filtering event types is not yet implemented. -> implemented
* [x] Is archiving still possible? Is it necessary? Can this possibly be implemented with tags? -> for DCB: no
* [x] The Projection API isn't fully developed yet. Returning QueryComponent instead of tags directly? Apply should be converted to Attributes. -> done


src/DCB/EventRouter.php
```diff
@@ -0,0 +1,172 @@
...
+use ReflectionUnionType;
+
+use function array_key_exists;
+use function array_keys;
+use function array_map;
+use function array_merge;
+use function class_exists;
+
+/** @experimental */
+trait EventRouter
```
DanielBadura: I think this might be misleading from the naming pov, as this is not really a router implementation in a classical way. Here we have mixed in apply logic (maybe this can be called routing) and query building / filtering. That being said: I come not with a better name 🤣 

Maybe we should add a `require-implements` for the `Projection` interface here.
Ref: https://phpstan.org/writing-php-code/phpdocs-basics#enforcing-implementing-an-interface-for-traits

src/Metadata/PropertyType.php
```diff
@@ -0,0 +1,58 @@
...
+
+use ReflectionParameter;
+use Symfony\Component\TypeInfo\Type;
+use Symfony\Component\TypeInfo\Type\ObjectType;
+use Symfony\Component\TypeInfo\Type\UnionType;
+use Symfony\Component\TypeInfo\TypeResolver\TypeResolver;
+
+use function array_map;
+
+class PropertyType
```
DanielBadura: Where do we use it? final and experimental missing, no?

src/Stringable.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Patchlevel\EventSourcing;
+
+use Patchlevel\EventSourcing\Serializer\Normalizer\StringableNormalizer;
+
+#[StringableNormalizer]
+interface Stringable
```
DanielBadura: Not sure if we should mark this also as experimental since this is only introduced because of dcb

tests/DbalManager.php
```diff
@@ -45,6 +46,14 @@ public static function createConnection(string $dbName = self::DEFAULT_DB_NAME):
         $databases = $schemaManager->listDatabases();
 
         if (in_array($dbName, $databases, true)) {
+            if ($tempConnection->getDatabasePlatform() instanceof PostgreSQLPlatform) {
+                $tempConnection->executeStatement("
+                    SELECT pg_terminate_backend(pid)
+                    FROM pg_stat_activity
+                    WHERE datname = '{$dbName}';
+                ");
+            }
```
DanielBadura, line 55: Could you explain this? :D

src/Metadata/PropertyType.php
DavidBadura: removed

src/Stringable.php
DavidBadura: 👍

src/DCB/EventRouter.php
DavidBadura: i added `require-implements`, I still have to think about the name.

tests/DbalManager.php
DavidBadura: Locally, I keep having the problem that the Postgres database ca not be deleted during tests. The reason is that an active connection supposedly still exists. This does not happen on the CI. I have spent a long time debugging and researching, but it seems to be a race condition. To work around this, it is recommended to kill the active connection before attempting to delete the database. That is what I did here.

Makefile
```diff
@@ -74,7 +74,7 @@ test: phpunit
 
 .PHONY: benchmark
 benchmark: vendor                                                               ## run benchmarks
-	DB_URL=sqlite3:///:memory: php -d memory_limit=512M vendor/bin/phpbench run tests/Benchmark --report=default
+	DB_URL=sqlite3:///:memory: php -d memory_limit=512M vendor/bin/phpbench run tests/Benchmark --report=default --filter=benchAppend100Events
```
DanielBadura: Filter needs to be removed?

docs/pages/dynamic_consistency_boundary.md
```diff
@@ -0,0 +1,454 @@
...
+$commandBus->dispatch(new CreateHotel($hotelId, 'HOTEL'));
+$commandBus->dispatch(new CheckIn($hotelId, 'David'));
+$commandBus->dispatch(new CheckIn($hotelId, 'Daniel'));
+$commandBus->dispatch(new CheckOut($hotelId, 'David'));
+```
+## Conclusion
+
+We've seen how to use DCB to make decisions consistently.
+In this example we skipped the subscription part,
+but you can add it by following the [Getting Started](./getting_started.md) section.
```
DanielBadura: I would write something like: As it is the same as before with aggregates

docs/pages/dynamic_consistency_boundary.md
```diff
@@ -0,0 +1,454 @@
...
+        $this->eventAppender->append([
+            new HotelCreated($command->hotelId, $command->hotelName),
+        ], $state->appendCondition);
+    }
+}
+```
+!!! note
+
+    Handlers build a Decision Model from the projections and then append events with an optimistic AppendCondition. 
+    If any relevant event arrives between read and write, the append fails and you can retry.
```
DanielBadura, line 303: maybe also add here an information about the phpstan extension and that there is then type safety when accessing the array

src/Store/Query.php
```diff
@@ -0,0 +1,53 @@
...
+
+    public function __construct(
+        SubQuery ...$subQueries,
+    ) {
+        $this->subQueries = array_values($subQueries);
+    }
+
+    public function add(SubQuery $subQuery): self
+    {
+        return new self(...array_merge($this->subQueries, [$subQuery]));
```
DanielBadura:
```suggestion
        return new self([...$this->subQueries, $subQuery]);
```
Should this be sufficient and maybe a little bit more performant?

tests/Benchmark/SimpleSetupTaggableStoreBench.php
```diff
@@ -0,0 +1,147 @@
...
+                    null,
+                ),
+            ),
+        ];
+
+        $this->store->append($messages);
+    }
+
+    #[Bench\Revs(10)]
+    public function benchAppend100Events(): void
```
DanielBadura, line 131: why 100 and not e.g. 1000 to be on the same amount as our other tests?

DavidBadura: Because unfortunately we are reaching the limits of how much data we can write at the same time. Unfortunately, we cannot batch here.

DanielBadura (approved):

PR #771, opened 2025-09-02
add identifier component

DavidBadura:
We will have different types of identifiers in the future, not just "aggregate root id." Therefore, it makes sense to create a general interface for them.

In addition, the return type of `fromString` has changed from `self` to `static`.


DavidBadura:
@DanielBadura Should we keep `AggregateRootId` and inherit from `Identifier`?

DanielBadura:
I'm not sure. This would make the BC-Break minimal for the users, but do we gain something else from it? Right now, we don't have anything where we would want to differntiate between an `Identifier` or `AggregateRootId`, do we?

DavidBadura:
Exactly, for BC, and to be more specific, an `AggregateRootId` is required here. But I do not think that really makes sense.

DanielBadura (approved):

PR #777, opened 2025-09-09
upgrade hydrator 2.0

### LINKED ISSUE #815: Support Upcasting of Headers (or Messages)
Right now, upcasting is limited to the event data itself. We have a case now, where we would like to upcast headers (add tenant information). As we are using our own serializers, we added the support for this in our own headers serializer.

Wondering though if it makes sense to add this as a general feature - or just pass the entire `Message` to upcasters instead? The later would keep the implementation cleaner (no separation between different kinds of upcasters).

DavidBadura:


docs/pages/index.md
```diff
@@ -11,7 +11,7 @@ powered by the reliable Doctrine ecosystem and focused on developer experience.
 * Automatic [snapshot](snapshots.md)-system to boost your performance
 * [Split](split_stream.md) big aggregates into multiple streams
 * Versioned and managed lifecycle of [subscriptions](subscription.md) like projections and processors
-* Safe usage of [Personal Data](personal_data.md) with crypto-shredding
+* Safe usage of [Personal Data](sensitive_data.md) with crypto-shredding
```
DanielBadura:
```suggestion
* Safe usage of [Sensitive Data](sensitive_data.md), like personal data, with crypto-shredding
```

docs/pages/normalizer.md
```diff
@@ -459,4 +459,4 @@ final class DTO
 * [How to define aggregates](aggregate.md)
 * [How to define events](events.md)
 * [How to snapshot aggregates](snapshots.md)
-* [How to work with personal data](personal_data.md)
+* [How to work with personal data](sensitive_data.md)
```
DanielBadura:
```suggestion
* [How to work with sensitive data](sensitive_data.md)
```

docs/pages/snapshots.md
```diff
@@ -261,4 +261,4 @@ And if the version is no longer correct and the snapshot is therefore invalid, t
 * [How to define aggregates](aggregate.md)
 * [How to store and load aggregates](repository.md)
 * [How to split streams](split_stream.md)
-* [How to work with personal data](personal_data.md)
+* [How to work with personal data](sensitive_data.md)
```
DanielBadura:
```suggestion
* [How to work with sensitive data](sensitive_data.md)
```

DanielBadura:
This could also land in 3.x, no?

DanielBadura:
Ok, now this cannot land anymore in 3.x

DavidBadura:
I would first look at what the optimal version for 4.x would look like and then see if it can be implemented in 3.x.

docs/sensitive-data.md
```diff
@@ -209,10 +217,10 @@ To remove personal data, you can either remove the key manually or do it with a
 use Patchlevel\EventSourcing\Attribute\Processor;
 use Patchlevel\EventSourcing\Attribute\Subscribe;
 use Patchlevel\EventSourcing\Message\Message;
-use Patchlevel\Hydrator\Cryptography\Store\CipherKeyStore;
+use Patchlevel\Hydrator\Extension\Cryptography\Store\CipherKeyStore;
 
 #[Processor('delete_personal_data')]
-final class DeletePersonalDataProcessor
+final class DeleteSensitiveDataProcessor
```
DanielBadura, line 223: should be consistent

docs/upcasting.md
```diff
@@ -3,93 +3,123 @@
...
+);
 ```
-:::tip
-Events can also have [aliases](events.md#alias). This is usually sufficient.
 :::
+## Adjust event name
+
+Renaming an event through an upcaster is no longer supported, because the event class is
+already resolved from the stored event name before the payload is upcasted.
+Use [event aliases](events.md#alias) instead.
```
DanielBadura, line 69: > Renaming an event through an upcaster is no longer supported,

This phrasing is better suited in the upgrade docs. Here in the actual 4.x docs, we should not say that its is no longer supported. Instead we should hint the user how he could solve it and link to the right direction. So I think a note how to get rid of old event names form the store should be added, if we keep this section here: recreate the store (which even does not need any translator)

docs/UPGRADE-4.0.md
```diff
@@ -327,3 +327,277 @@ and replaced with the following headers:
...
+```
+after:
+
+```php
+$cipherKeyStore->removeWithSubjectId($subjectId);
+```
+
+:::danger
+The key table layout changed (`crypto_keys` -> `cryptography_keys` with new columns).
+Existing keys must be migrated, otherwise stored sensitive data can no longer be decrypted.
```
DanielBadura, line 595: We should add a note how to migrate

DavidBadura:
Superseded by #900, which does the hydrator 2.0 upgrade on top of the current 4.0.x. The relevant parts from here were taken over: the event name and class are passed as context to the hydrator, the headers serializer factories accept a custom hydrator, the upgrade guide covers the moved cryptography attributes, and the personal data page was renamed to sensitive data. The DefaultSnapshotStore constructor change was left out since it is unrelated to the hydrator upgrade.

PR #779, opened 2025-09-16
add first previous exception in error detected exception for better stack trace

DavidBadura:


DanielBadura:
Hm I'm not sure why this would improve debugging as we are loosing information here?

DavidBadura:
We are not losing any information here. All information is still present in the exception. What has been improved here is the stack trace, which previously ended with this exception because no previous exception was used. Instead, all exception messages were displayed, but they often were not helpful because it was unclear where they came from exactly. And instead of showing all messages with little information, the first one is displayed with maximum information.

Before:

```
Patchlevel\EventSourcing\Subscription\Engine\ErrorDetected: Error in subscription engine detected.
Subscription push: Failed to send an update.

Stack trace Ended
```

After:
```
Patchlevel\EventSourcing\Subscription\Engine\ErrorDetected: 1 error(s) in subscription engine detected. First error is in "push" subscription: Failed to send an update.

+ Stack trace of first exception
```

src/Subscription/Engine/ErrorDetected.php
```diff
@@ -16,15 +15,15 @@ final class ErrorDetected extends RuntimeException
...
+        parent::__construct(
+            sprintf(
+                '%s error(s) in subscription engine detected. First error in "%s" subscription: %s',
+                count($errors),
+                $errors[0]->subscriptionId,
+                $errors[0]->message,
             ),
-            $errors,
+            0,
+            $errors[0]->throwable,
```
DanielBadura: maybe use here a named parameter to mitigate the error code? Or add this to a test 🙈

DanielBadura (approved):

PR #807, opened 2026-01-13
add subscription cleanup feature

### LINKED ISSUE #806: Detached subscriptions can’t be torn down after renaming/versioning the projection id (old subscriber not found) — clarify recommended workflow / docs
When I change a subscription/projection id to version it (e.g. abc_1 → abc_2), the previous subscription (abc_1) becomes detached (e.g. after running event-sourcing:subscription:boot --setup).
However, when I try to clean up using:
```bash
bin/console event-sourcing:subscription:teardown
```

the old one cannot be removed anymore because the subscriber no longer exists in the current codebase:
```
Subscriber for "abc_1" to teardown not found, skipped.
```

Result: outdated projection tables remain in the database and accumulate over time.

A bit more context to our setup:
* We deploy by replacing containers (Ansible).
* We already run `bin/console event-sourcing:subscription:boot --setup` in the container init.
* We do not have a guaranteed “teardown with the old container/code” lifecycle step (like a Kubernetes preStop), so cleanup of old versions is not straightforward. But might be an option to take a look at, if there are no other solution

DavidBadura:
closes https://github.com/patchlevel/event-sourcing/issues/806

Add an alternative way to clean up subscriptions without requiring the subscriber to still exist in the code.

When creating a subscription, the engine checks if a cleanup method exists. If so, it is called, and the cleanup tasks, along with other metadata, are stored in the subscription store. When the subscriber is removed, instead of executing the teardown method, the tasks from the store are used, and the corresponding handlers are called. This eliminates the need for the subscriber code itself. By default, we have added two tasks for Doctrine/dbal: `DropTableTask` and `DropIndexTask`. It is extensible. Users can also create their own tasks and handlers, for example, if they are using MongoDB.


src/Subscription/Cleanup/Dbal/DbalCleanupHandler.php
```diff
@@ -0,0 +1,33 @@
...
+    public function __invoke(object $task): void
+    {
+        switch ($task::class) {
+            case DropTableTask::class:
+                $this->connection->createSchemaManager()->dropTable($task->table);
+                break;
+            case DropIndexTask::class:
+                $this->connection->createSchemaManager()->dropIndex($task->index, $task->table);
+                break;
+        }
```
DanielBadura: Maybe we should return early here & throw an exception if no case exist or add a default which throws? Just to be sure that this method and the supports are never out of sync?

DanielBadura (approved):
LGTM, maybe add some more tests as we have on the diff only a coverage of 40%. That is pretty low

DanielBadura (approved):

PR #811, opened 2026-02-10
Add paragraph on implementing custom resolvers for subscribers

fritz-gerneth:


docs/pages/subscription.md
```diff
@@ -293,6 +293,12 @@ final class DoStuffSubscriber
     }
 }
 ```
+
+##### Custom Resolvers
+
+You can provide your own argument resolvers by implementing the `ArgumentResolver` interface. 
+This can be useful to provide direct access to custom headers or any other third party.
```
DanielBadura:
```suggestion
You can provide your own argument resolvers by implementing the `ArgumentResolver` interface.
This can be useful for providing direct access to custom headers or other third party headers.
```

DanielBadura (approved):
just a little wording nitpick, thanks!

docs/pages/subscription.md
fritz-gerneth: This is not limited to headers but any kind of data / service (these should be injected through the constructor though). Added this as `data`

DanielBadura: That's true, as e.g. the `Lookup` is also handled via an `ArgumentResolver`.

DavidBadura (approved):

PR #812, opened 2026-02-10
Add nullable information to Subscriber ArgumentMetadata

fritz-gerneth:
Not sure if to make this argument mandatory, would be a breaking change if this is considered public API. Do not like default here either though (no reasonable default value)

### DIFF
--- a/src/Metadata/Subscriber/ArgumentMetadata.php
+++ b/src/Metadata/Subscriber/ArgumentMetadata.php
@@ -9,6 +9,7 @@ final class ArgumentMetadata
     public function __construct(
         public readonly string $name,
         public readonly string $type,
+        public readonly bool $allowsNull = false,
     ) {
     }
 }
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -463,7 +463,7 @@ parameters:
 			path: tests/Unit/Message/Translator/ReplaceEventTranslatorTest.php
 
 		-
-			message: '#^Method class@anonymous/tests/Unit/Metadata/Subscriber/AttributeSubscriberMetadataFactoryTest\.php\:212\:\:profileVisited\(\) has parameter \$message with no type specified\.$#'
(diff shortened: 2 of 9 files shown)


src/Metadata/Subscriber/ArgumentMetadata.php
```diff
@@ -9,6 +9,7 @@ final class ArgumentMetadata
     public function __construct(
         public readonly string $name,
         public readonly string $type,
+        public readonly bool $allowsNull,
```
DanielBadura:
```suggestion
        public readonly bool $allowsNull = false,
```

The default should be `false`, as this is the current situation pre this patch. And yes, this is public api. We currently have only one class which is marked internal and thereby is not public api :D

DanielBadura (changes requested):

src/Metadata/Subscriber/ArgumentMetadata.php
fritz-gerneth: How about making it default null instead? The reason I opted against null is that this can be false information without means to detect this (not passed does not mean not nullable). Having it null would suggest 'unknown'. But that puts more burden on users to also check for null of course :)

DanielBadura: But what would the behaviour be, if `$allowsNull` is `null`? Either it supports nullable values (`true`) or not (`false`), i cannot thing of a third option here 🤔

fritz-gerneth: Behavior would be left to the recipient of the `ArgumentMetadata`. This class is only about the information we have, `null` meaning "we do not have it". 

But since nullable arguments were not supported before:
- they probably were either not nullable before
- or if nullable, they were never called with null

So I do not really have a strong opinion about this either way :)

DanielBadura: I would go with `false` as the default value as this was not supported before it is imho the right choice.

DanielBadura (approved):

DanielBadura:
Thank you!

PR #813, opened 2026-02-11
allow to refresh subscriptions

### LINKED ISSUE #747: Allow to change subscription `group` & `run mode` afterwards

DavidBadura:
fix: https://github.com/patchlevel/event-sourcing/issues/747

Allow the refreshing of subscriptions, such as run-mode and group. If the PR https://github.com/patchlevel/event-sourcing/pull/807 is merged, cleanup tasks will also be added here.

### DIFF
--- a/src/Subscription/Engine/SubscriptionRefreshable.php
+++ b/src/Subscription/Engine/SubscriptionRefreshable.php
@@ -1,0 +1,8 @@
+<?php
+
+namespace Patchlevel\EventSourcing\Subscription\Engine;
+
+interface SubscriptionRefreshable
+{
+    public function refreshSubscriptions(SubscriptionEngineCriteria|null $criteria = null): Result;
+}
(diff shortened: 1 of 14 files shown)


src/Subscription/Engine/SubscriptionRefreshable.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Patchlevel\EventSourcing\Subscription\Engine;
+
+interface SubscriptionRefreshable
```
DanielBadura: This sounds like the interface should go on a subsription and not on the engine. Maybe something like `CanRefreshSubscriptions`?

Also, this interface would merge into `SubscriptionEngine` in `4.0.x`?

DavidBadura: Yes, I would include it in the SubscriptionEngine interface in version 4.0.

DanielBadura (approved):

PR #814, opened 2026-02-14
use type info in argument resolver

DavidBadura:
Use `symfony/type-info` for the subscriber argument resolver. This is a further step to remove custom type logic and thus our limitations.


tests/Unit/Metadata/Subscriber/AttributeSubscriberMetadataFactoryTest.php
```diff
@@ -245,38 +245,6 @@ public function profileVisited($message): void
         $metadataFactory->metadata($subscriber::class);
     }
 
-    public function testUnionTypeNotSupported(): void
```
DanielBadura, line 248: if we are now supporting these cases, we should cover them, no?

DavidBadura: In my opinion, that doesn't make sense, since we neither actively support it nor don't. We add the argument type to the argument metadata (this is tested), and the argument resolver decides whether something is supported (this is also tested). `type info` supports union types as well as collections, array shapes, etc. (we haven't tested this, we don't test the functionality of external code).

src/Metadata/Subscriber/AttributeSubscriberMetadataFactory.php
```diff
@@ -171,18 +178,9 @@ private function subscribeMethod(ReflectionMethod $method): SubscribeMethodMetad
                 );
             }
 
-            if (!$type instanceof ReflectionNamedType) {
```
DanielBadura, line 174: Here we could add e.g. a check for union / intersection types, if we dont want to support it:

```php
$resolvedType = $this->typeResolver->resolve($type);

if ($resolvedType instanceof UnionType || $resolvedType instanceof IntercetionType) {
    throw new NotSupported();
}

            $arguments[] = new ArgumentMetadata(
                $parameter->getName(),
                $resolvedType,
            );
```

tests/Unit/Metadata/Subscriber/AttributeSubscriberMetadataFactoryTest.php
DanielBadura: I don't want to test external code. I want to test, if our code can now support these cases. We also could add checks to disallow these cases explictly, if we don't want to support them.

I just want clarity of what the user is capable of doing in our subcriptions as this is a user facing API. 

I think we should either:
* Disallow these types (see other comment), keep the tests and document it.
* Allow these types, adjust tests and document it 

Maybe not part as a unit test here, but more as an integration test?

As this should now be possible which was not before if I'm not mistaken:
```php
            #[Subscribe(ProfileVisited::class)]
            #[Subscribe(ProfileCreated::class)]
            public function profileVisited(ProfileVisited|ProfileCreated $event): void
            {
            }
```

src/Metadata/Subscriber/AttributeSubscriberMetadataFactory.php
DavidBadura: I don't understand why I should do that. Why do we want to prevent that?

DanielBadura: No. I dont want that. It was only one option from my pov.

DanielBadura (approved):

PR #821, opened 2026-02-19
allow multiple handlers, union types and inheritance for commands

DavidBadura:


docs/pages/command_bus.md
```diff
@@ -45,6 +45,59 @@ final class CreateProfileHandler
...
+```php
+use Patchlevel\EventSourcing\Attribute\Handle;
+
+final class CreateProfileHandler
+{
+    #[Handle]
+    public function __invoke(CommandInterface $command): void
+    {
+        // handle all commands that implement CommandInterface
+    }
```
DanielBadura, line 98: Just a nit: 
imho this example is a bad one, as you should probably never have this kind of handler which just handles all commands. 

We should choose an example which is more reasonable to argue with. But this should not block this PR

DanielBadura (approved):

docs/pages/command_bus.md
DavidBadura: Obviously, it is best not to define a handler that acts on all commands, but I still think the example is good because it shows exactly what it does.

PR #822, opened 2026-02-20
add auto initialize aggregate feature

### LINKED ISSUE #769: Allow implicit creation of Aggregates or defintition of custom repositories
We sometimes have Aggregates that do not have an explicit `create` event, they just come into existence by whatever event is recorded first (often this is some ID managed outside the context of that particular Aggregate, e.g. some external IDs like users coming from external IDPs). 

Right now we have to create a custom repository with a `load` function somewhat like this:
```
    public function load(Uuid $aggregateId): MyAggregate
    {
        return $this->repository->has($aggregateId)
            ? $this->repository->load($aggregateId)
            : MyAggregate::create($aggregateId);
    }
```

As we cannot define our own repository (easily) to be used by the `DefaultRepositoryMananger`, we loose the benefits of defining `#[Handler]` directly in the aggregate and all the loading/saving coming with it. Instead we have to go manual and create custom handler classes for all commands of this aggregate (using Symfony):
```
    #[AsMessageHandler]
    public function myCommandHanlder(MyCommand $command): void
    {
        $aggregate = $this->repository->load($command->id);
        $aggregate->myCommand($command);
        $this->repository->save($aggregate);
    }
```

As I see it, this could be solved by two means:
- Allow extension of the default repository class, and extend the `Aggregate` attribute to allow specification of the repository class to use. This is what  Doctrine does with their entity repositories: `#[Entity(repositoryClass: MyRepository::class)` and would tie nicely with the repository manager. E.g. `$repositoryManager->get(MyAggregate::class);` would return the custom implementation specified in the attribute. Note that doctrine bundle also just uses a proxy for the actual [implementation](https://github.com/doctrine/DoctrineBundle/blob/2.15.x/src/Repository/ServiceEntityRepositoryProxy.php) for convenience of passing arguments.
- The alternative would be to add an optional argument `allow_implicit_create` argument to the `Aggregate` attribute. The `DefaultRepository::load` method already knows about the `Id` property, so it could instantiate a new aggregate instance and set the ID to return an empty aggregate root. 

Personally I prefer the first option - it leaves more room for extension than just pilling flags on the `Aggregate` that inherently should affect the repository, not the aggregate. Using the SymfonyBundle, this currently would entail to overwrite the `DefaultRepositoryManager`.

 I guess this would mean to:
- extend the `AggregateRootMetadata` with an additional property for the repository class
- extend the `DefaultRepositoryManager` to check for the property instead of instantiating the default repository
- find some pattern to hide away all the extra parameters the `DefaultRepository` does need

This is the current implementation we use by decorating the `RepositoryManager` (simply using a static array of aggregates to handle different):
```
#[AsDecorator(decorates: DefaultRepositoryManager::class)]
class ImplicitCreateRepositoryManager implements RepositoryManager
{
    private static array $implicitlyCreatedAggregates = [
        FeatureFlagOverwrite::class => true,
    ];

    public function __construct(
        private readonly RepositoryManager $repositoryManager,
        private readonly AggregateRootMetadataFactory $metadataFactory,
    ) {
    }

    public function get(string $aggregateClass): Repository
    {
        $repository = $this->repositoryManager->get($aggregateClass);

        return array_key_exists($aggregateClass, self::$implicitlyCreatedAggregates)
            ? new ImplicitCreateRepository($repository, $this->metadataFactory->metadata($aggregateClass))
            : $repository;
    }
}

class ImplicitCreateRepository implements Repository
{
    public function __construct(
        private readonly Repository $repository,
        private readonly AggregateRootMetadata $metadata,
    ) {
    }

    public function load(AggregateRootId $id): AggregateRoot
    {
        if ($this->repository->has($id)) {
            return $this->repository->load($id);
        }

        $reflectionClass = new ReflectionClass($this->metadata->className);
        
        $aggregate = $reflectionClass->newInstanceWithoutConstructor();
        $reflectionClass
            ->getProperty($this->metadata->idProperty)
            ->setValue($aggregate, $id);

        return $aggregate;
    }

    public function has(AggregateRootId $id): bool
    {
        return $this->repository->has($id);
    }

    public function save(AggregateRoot $aggregate): void
    {
        $this->repository->save($aggregate);
    }
}
```

DavidBadura:
fix #769

### DIFF
--- a/docs/pages/aggregate.md
+++ b/docs/pages/aggregate.md
@@ -876,6 +876,53 @@ final class Order extends BasicAggregateRoot
     }
 }
 ```
+
+## Auto Initialize
+
+??? example "Experimental"
+
+    This feature is still experimental and may change in the future.
+    Use it with caution.
+
+Sometimes you want to be able to access an aggregate even if it has not yet been created in the system. 
+In this case, the aggregate should be automatically initialized if it cannot be found in the store. 
+To achieve this, the aggregate must mark the initialization method with the `AutoInitialize` attribute.
+The method must be static, receives the aggregate ID as an argument and must return an instance of the aggregate.
+
+```php
+use Patchlevel\EventSourcing\Aggregate\BasicAggregateRoot;
+use Patchlevel\EventSourcing\Aggregate\Uuid;
+use Patchlevel\EventSourcing\Attribute\Aggregate;
+use Patchlevel\EventSourcing\Attribute\Apply;
+use Patchlevel\EventSourcing\Attribute\AutoInitialize;
+use Patchlevel\EventSourcing\Attribute\Id;
+
+#[Aggregate('profile')]
+final class Profile extends BasicAggregateRoot
+{
+    #[Id]
+    private Uuid $id;
+
+    #[AutoInitialize]
+    public static function initialize(Uuid $id): static
+    {
+        $self = new static();
+        $self->recordThat(new ProfileCreated($id));
+
+        return $self;
+    }
+
+    #[Apply]
+    public function applyProfileCreated(ProfileCreated $event): void
+    {
+        $this->id = $event->id;
+    }
+}
+```
+!!! note
+
+    Recording events in the `initialize` method is optional but recommended.
+
 ## Aggregate Root Registry
 
 The library needs to know about all aggregates so that the correct aggregate class is used to load from the database.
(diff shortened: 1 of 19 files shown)


DanielBadura (approved):
@fritz-gerneth this should also work for your usecase, right?

PR #832, opened 2026-03-16
Add layer tests with `phpat` based on `deptrac` layers, remove `deptrac`

DanielBadura:
This is the next step into removing deptrac in favor of phpat.

These 3 violations, which are in the deptrace baseline, are not covered right now:
```yaml
    Patchlevel\EventSourcing\Aggregate\CustomId:
      - Patchlevel\EventSourcing\Serializer\Normalizer\IdNormalizer
    Patchlevel\EventSourcing\Aggregate\Uuid:
      - Patchlevel\EventSourcing\Serializer\Normalizer\IdNormalizer
    Patchlevel\EventSourcing\Attribute\Stream:
      - Patchlevel\EventSourcing\Aggregate\AggregateRoot
```

These 2 are false positives from deptrac imho:
```yaml
    Patchlevel\EventSourcing\Aggregate\CustomId:
      - Patchlevel\EventSourcing\Serializer\Normalizer\IdNormalizer
    Patchlevel\EventSourcing\Aggregate\Uuid:
      - Patchlevel\EventSourcing\Serializer\Normalizer\IdNormalizer
```

As the `AggregateRootId` interface as a dependency on the attribute `Patchlevel\EventSourcing\Serializer\Normalizer\IdNormalizer` and not the concrete classes. So this is a good that these 2 are not picked up.

```yaml
    Patchlevel\EventSourcing\Attribute\Stream:
      - Patchlevel\EventSourcing\Aggregate\AggregateRoot
```

Is due the dependecy is coming from a union type inside the phpdoc. As soon as I change `/** @param string|class-string<AggregateRoot> $name */` to `/** @param class-string<AggregateRoot> $name */` it gets picked up correctly:
```
 ------ ----------------------------------------------------------------------------------------------------------------- 
  Line   src/Attribute/Stream.php                                                                                         
 ------ ----------------------------------------------------------------------------------------------------------------- 
  14     Patchlevel\EventSourcing\Attribute\Stream should not depend on Patchlevel\EventSourcing\Aggregate\AggregateRoot  
         🪪  phpat.testAttributeCanOnlyDependOnAllowedLayers                                                              
 ------ ----------------------------------------------------------------------------------------------------------------- 
```

So this is a bug on phpat side.

With this change, we can remove deptrac as an dependency and still have our layers protected. WDYT @DavidBadura ?


DavidBadura (approved):

DanielBadura:
@DavidBadura not sure what the error is in the BC-Check. But it's for sure unrelated :D

DavidBadura (approved):
LGTM 👍

PR #834, opened 2026-04-01
dbal cleanup handler: check if a table or index exists before deleting it

DavidBadura:


phpstan-baseline.neon
```diff
@@ -126,6 +126,12 @@ parameters:
 			count: 1
 			path: src/Store/StreamDoctrineDbalStoreStream.php
 
+		-
+			message: '#^Parameter \#1 \$tableName of method Doctrine\\DBAL\\Schema\\AbstractSchemaManager\<Doctrine\\DBAL\\Platforms\\AbstractPlatform\>\:\:introspectTableIndexesByUnquotedName\(\) expects non\-empty\-string, string given\.$#'
```
DanielBadura: shouldnt we adapt the type? as this makes perferctly sense to me

DanielBadura (approved):

PR #846, opened 2026-04-21
Add PSR-Container

DanielBadura:
Adds a PSR based container builder helper

- [x] Schema creation wiring
- [x] Decorator implementation missing (e.g. ReadOnlyStore)


composer.json
```diff
@@ -53,6 +53,7 @@
     "infection/infection": "^0.32.0",
     "league/commonmark": "^2.6.1",
     "patchlevel/coding-standard": "^1.3.0",
+    "patchlevel/event-sourcing-bundle": "^3.16",
```
DanielBadura: to be removed

PR #847, opened 2026-04-22
refactor stream

### LINKED ISSUE #729: Simplify Stream API
The Stream interface currently has two methods, `position()` and `index()`. These are difficult to implement and must be nullable (if the stream isn't already being iterated). The API should be simplified: Both methods will be removed. You can track the current position yourself. The index will be the array key in the future.

DavidBadura:
fix #729

The index and position remain unchanged. However, instead of having to implement the stream yourself, there is now a generic implementation available for use.


src/Subscription/Engine/EventFilteredStoreMessageLoader.php
```diff
@@ -26,9 +26,13 @@ public function __construct(
     }
 
     /** @param list<Subscription> $subscriptions */
-    public function load(int $startIndex, array $subscriptions): Stream
+    public function load(int|null $startIndex, array $subscriptions): Stream
```
DanielBadura, line 29: Why are we opening the type here?

src/Subscription/Store/DoctrineSubscriptionStore.php
```diff
@@ -230,7 +230,7 @@ public function configureSchema(Schema $schema, Connection $connection): void
             ->setLength(16)
             ->setNotnull(true);
         $table->addColumn('position', Types::INTEGER)
-            ->setNotnull(true);
+            ->setNotnull(false);
```
DanielBadura, line 233: Ah, ok so message loader needs `null` because of this change here. But why are we doing this here?

DavidBadura: Previously, we used "0" as the starting value for loading messages because the stream index always started at "1" (due to auto-increment). Since we now also allow the index "0" and even (theoretically) negative numbers, we can no longer determine the starting value.

Furthermore, "0" was more of a magic number in this context. "0" meant that the subscriber hadn't processed any messages yet. But now, in combination with an index that can start at "0", the position "0" means that the message with index "0" has been processed.

It makes much more sense to use "null" because it perfectly covers the case where the subscriber has no position because they haven't processed any messages yet.

DanielBadura: Ok, thanks!

DanielBadura (approved):

PR #854, opened 2026-06-01
Improve docs around worker

Brammm:
While setting up patchlevel myself, I found the docs a little lacking around the Message Loader stuff and the process getting the worker set up. I think it can still be improved, it's mainly my own lack of knowledge about the subscription commands holding me back.

### DIFF
--- a/docs/pages/cli.md
+++ b/docs/pages/cli.md
@@ -43,7 +43,7 @@ To manage your subscriptions there are the following cli commands.
 !!! note
 
     You can find out more about subscriptions [here](subscription.md).
-    
+
 ## Inspector commands
 
 The inspector is a tool to inspect the event streams.
--- a/docs/pages/getting_started.md
+++ b/docs/pages/getting_started.md
@@ -335,6 +335,12 @@ $hotelRepository = $repositoryManager->get(Hotel::class);
     `DefaultRepositoryManager` and a worker to run the Subscription Engine.
     
     Learn more [here](subscription.md).
+
+!!! warning
+    From version 4.x onward, a MessageLoader will be required for the SubscriptionEngine 
+    instead of a Store implementation. 
+    Read more about it [here](./subscription.md#message-loader). In current versions, 
+    you can still provide a Store, the SubscriptionEngine will wrap it internally. 
     
 ## Database setup
 
--- a/docs/pages/subscription.md
+++ b/docs/pages/subscription.md
@@ -838,13 +838,28 @@ In order for the subscription engine to be able to do its work, you have to asse
 ### Message Loader
 
 The subscription engine needs a message loader to load the messages.
-We provide two implementations by default.
+We provide three implementations by default.
 Which one has a better performance depends on the use case.
 
 !!! tip
 
-    We recommend the `GapResolverStoreMessageLoader` as it handles gaps in the stream.
+    We recommend the recent `GapResolverStoreMessageLoader` as it handles gaps in the stream.
+
+!!! note
+
+    If you followed the confiration in the [Getting Started](./getting_started.md#configuration) section,
+    the Message Loader implementation replaces the $eventStore in the DefaultSubscriptionEngine.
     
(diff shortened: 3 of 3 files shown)


docs/pages/cli.md
```diff
@@ -43,7 +43,16 @@ To manage your subscriptions there are the following cli commands.
...
-    
+
+## Basic workflow for the worker
+
+Use `event-sourcing:subscription:boot --setup` to first run the setup of any new subscriptions
+and immediately boot them.
+
+The `event-sourcing:subscription:run` command will continue to run and process new events
+until the process is killed. After adding a new subscriber and booting it, you should 
+restart the `run` command.
```
DanielBadura: I think it is better placed in the subscription section. We already have a [usage docs subsection](https://event-sourcing.patchlevel.io/latest/subscription/#usage), but it is referencing how to interact with the engine directly, which is imho the rarer case then to call the cli commands. Alo maybe it should more visible, like moving it up in the subscription section? 

I think it is definitely a nice addtion to the docs. @DavidBadura any suggestions from your side as you made most of the docs structure?

DanielBadura (commented):
The other doc changes are also looking good from my side 👍

docs/pages/subscription.md
```diff
@@ -838,13 +838,28 @@ In order for the subscription engine to be able to do its work, you have to asse
...
     
+    ```php
+        $messageLoader = // whatever implementation you choose
+        $engine = new DefaultSubscriptionEngine(
+    -       $eventStore,
+    +       $messageLoader,
+            $subscriptionStore,
+            $subscriberRepository,
+        );
+    ```
```
DavidBadura, line 861: I would prefer to adjust the "Getting Started" guide, since in `4.x`, only a `MessageLoader` can be passed.

docs/pages/cli.md
DavidBadura: I would also tend to move it in the SubscriptionEngine documentation.

Brammm: done

docs/pages/subscription.md
Brammm: Added a warning to getting started and link to subscription docs

docs/pages/getting_started.md
```diff
@@ -335,6 +335,12 @@ $hotelRepository = $repositoryManager->get(Hotel::class);
     `DefaultRepositoryManager` and a worker to run the Subscription Engine.
     
     Learn more [here](subscription.md).
+
+!!! warning
+    From version 4.x onward, a MessageLoader will be required for the SubscriptionEngine 
+    instead of a Store implementation. 
+    Read more about it [here](./subscription.md#message-loader). In current versions, 
+    you can still provide a Store, the SubscriptionEngine will wrap it internally.
```
DanielBadura, line 343: I think you misunderstood. Let's just update the getting started to use the message loader instead of adding a warning here.

```php
use Patchlevel\EventSourcing\Subscription\Engine\StoreMessageLoader;

$engine = new DefaultSubscriptionEngine(
    new StoreMessageLoader($eventStore),
    $subscriptionStore,
    $subscriberRepository,
);
```

docs/pages/subscription.md
DanielBadura, line 861: then this is also not needed

DanielBadura (changes requested):

docs/pages/getting_started.md
DanielBadura: If you open the PR that maintainers can edit it, I can also do the adjustment :)

DanielBadura:
@Brammm I created another PR which is adressing all points from here as we did launch a new docs website which led to some changes in the doc structure which i did not want you to resolve here. Also, i adjusted the getting started to use the message loader instead of adding some warnings in the docs. Thanks for the contribution!

Ref: https://github.com/patchlevel/event-sourcing/pull/858

PR #855, opened 2026-06-01
Add a `--force` param to skip the prompt

Brammm:
I personally like the option to skip a confirm prompt with a parameter. This allows me to use this command in some other tooling where I don't have the interactive prompt.


DanielBadura (approved):
LGTM, thanks @Brammm !

DavidBadura (approved):

PR #856, opened 2026-06-06
Fix some issues reported by mago

DanielBadura:


src/CommandBus/Handler/ParameterResolver.php
```diff
@@ -8,6 +8,6 @@
 
 interface ParameterResolver
 {
-    /** @return iterable<mixed> */
+    /** @return iterable<int, mixed> */
```
DanielBadura, line 11: this is technically a bc-break, but we are using the spread operator already on the result of `ParameterResolver::resolve`, which means it can be only be working with an `int` as the key. Also, we typed it like that in our concrete implementation https://github.com/patchlevel/event-sourcing/pull/856/changes#diff-fb59bb4d80adcd4ac904abb097ca2b9c8013a263a00117b936eaecc5f4d3a732L22

src/QueryBus/HandlerFinder.php
```diff
@@ -67,7 +67,6 @@ public static function findInClass(string $classString): iterable
                 );
             }
 
-            /** @var class-string $commandClass */
```
DanielBadura, line 70: `getClassName` is already typed as `T of class-string`

DanielBadura: phpstan seems to not really pick it up 🤔

tests/Unit/Subscription/ErrorContextTest.php
```diff
@@ -43,6 +43,7 @@ public function testErrorContext(): void
 
         $firstTrace = $error['trace'][0];
 
+        self::assertIsArray($firstTrace);
```
DanielBadura, line 46: phpstan seems to have a better type knowledge at this place then mago

PR #864, opened 2026-06-16
Add event emitter for subscriptions

DavidBadura:
Subscribers can emit events back into the store. If a subscribe method type-hints an `EventEmitter`, it gets injected and the subscriber can `emit()` events into its own `subscription_<id>` stream, or `linkTo()` an explicit stream. Another subscription can then subscribe to those events. Typical use case: a projection finishes updating and emits an event that a notification subscriber reacts to.

Emitting only happens during `run` by default; while booting the emitter is a noop so a replay doesn't duplicate events. `#[OverrideEventEmitting(true)]` also emits during boot, `false` suppresses it entirely.

When a subscription is removed the handlers dispatch `OnSubscriptionRemoved`. Registering the `RemoveSubscriptionStreamListener` on the event dispatcher removes the corresponding `subscription_<id>` stream from the store.


src/Attribute/OverrideEventEmitting.php
```diff
@@ -0,0 +1,16 @@
...
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_CLASS)]
+final class OverrideEventEmitting
+{
+    public function __construct(
+        public readonly bool $enabled,
+    ) {
+    }
```
DanielBadura: I think something like `#[DisableEventEmitting]` and `#[EmitDuringBoot]` would be better. It's more descriptive in what it is really doing and the scope here is clear.

src/Subscription/Subscriber/EventEmitter/DefaultEventEmitter.php
```diff
@@ -0,0 +1,41 @@
...
+
+declare(strict_types=1);
+
+namespace Patchlevel\EventSourcing\Subscription\Subscriber\EventEmitter;
+
+use Patchlevel\EventSourcing\Message\Message;
+use Patchlevel\EventSourcing\Store\Header\StreamNameHeader;
+use Patchlevel\EventSourcing\Store\Store;
+
+final class DefaultEventEmitter implements EventEmitter
```
DanielBadura:
```suggestion
final class StoreEventEmitter implements EventEmitter
```

DanielBadura (commented):
LGTM besides naming stuff, but we need some more unit tests as even the non-diff mutation test is failing

src/Attribute/OverrideEventEmitting.php
DavidBadura: And what happens if you set both attributes? I think it's better to keep it in a single attribute, because that's exactly what it does: it overrides the behavior.

DanielBadura (approved):

PR #865, opened 2026-06-16
Replace BatchableSubscriber interface with batch attributes

DavidBadura:
Batching for subscribers no longer relies on the `BatchableSubscriber` interface. The batch lifecycle is now declared with attributes, in line with the rest of the subscriber API which is already attribute and argument-resolver driven.

A subscriber opts into batching with `#[BatchBegin]`, `#[BatchFlush]`, `#[BatchRollback]` and optionally `#[BatchShouldFlush]`. The begin method may return a state object, which is injected into subscribe handlers via the `#[BatchState]` parameter attribute and passed to flush/rollback.


DanielBadura (approved):
You should adjust the infection.json msi values down to the current value with this PR

PR #867, opened 2026-06-19
Add `MissingHeaders` if no header found

### LINKED ISSUE #818: Make it possible to remove Header classes without crashes
Right now, if a header landed in the eventstore, you need to keep the code in place, even if you don't use the header anymore. If you remove the Header class currently the serialization process crashes.

Why is this important? There are cases where maybe a 3rd party bundle adds a header to you events and later on you remove the bundle from your dependencies. This will lead to a crash, as the header class could not be found in the header registry. This would rightnow force the user to one of the following solutions:
* Keep the bundle as a dep
* Copy the headers from the bundle and register them
* re-create the eventstore with a translator which removes the headers

I think it would be best to handle these cases more gracefully, as headers are mostly only metadata.

DanielBadura:
~It is kinda BC-Break, as the behaviour is changing greatly.~

Not sure if this is the right path, as the app could still try to load the original Header from the Message - and then it would still crash, but there we also have `hasHeader` so it is possible to handle it.

closes #818

### DIFF
--- a/src/Message/Serializer/DefaultHeadersSerializer.php
+++ b/src/Message/Serializer/DefaultHeadersSerializer.php
@@ -4,22 +4,30 @@ declare(strict_types=1);
 
 namespace Patchlevel\EventSourcing\Message\Serializer;
 
+use Patchlevel\EventSourcing\Message\MissingHeaders;
 use Patchlevel\EventSourcing\Metadata\Message\AttributeMessageHeaderRegistryFactory;
+use Patchlevel\EventSourcing\Metadata\Message\HeaderNameNotRegistered;
 use Patchlevel\EventSourcing\Metadata\Message\MessageHeaderRegistry;
 use Patchlevel\EventSourcing\Serializer\Encoder\Encoder;
 use Patchlevel\EventSourcing\Serializer\Encoder\JsonEncoder;
 use Patchlevel\Hydrator\Hydrator;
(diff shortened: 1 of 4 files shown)


DavidBadura (changes requested):
That should be optional. There are headers that are business-critical (e.g., the tenant header). if one of those isn't serialized for some reason, I think it should crash.

It would be best to be able to specify which headers are allowed to be "missing", so if a header is no longer present, you simply list it. Or you could specify `*` to allow all of them. Like `MissingApply`.

That way, it wouldn't be a breaking change.

src/Message/Serializer/DefaultHeadersSerializer.php
```diff
@@ -46,27 +59,48 @@ public function deserialize(string $string, array $options = []): array
...
-            );
+            try {
+                $headers[] = $this->hydrator->hydrate(
+                    $this->messageHeaderRegistry->headerClass($headerName),
+                    $headerPayload,
+                );
+            } catch (HeaderNameNotRegistered $exception) {
+                if (
+                    !in_array('*', $this->gracefulMissingHeaders, true)
+                    && !in_array($headerName, $this->gracefulMissingHeaders, true)
```
DavidBadura: This could be checked more efficiently. In the worst-case scenario, this logic runs for every message.

DavidBadura (approved):

DavidBadura (approved):

PR #873, opened 2026-06-30
cleanup docs

DavidBadura:


docs/UPGRADE-3.0.md
```diff
@@ -1,3 +1,6 @@
+---
+searchable: false
```
DanielBadura, line 2: Should this be added in general for all Upgrade docs, or just for this one as this is already old and most probably not needed anymore?

docs/introduction.md
```diff
@@ -1,3 +1,6 @@
+---
+searchable: false
```
DanielBadura, line 2: Just as a note: maybe we should add this in some kind of contribution docs. Most probably also a global "doc" page like the bc promise one? Or add an .md file in each repo?

DanielBadura (approved):

docs/UPGRADE-3.0.md
DavidBadura: None of the upgrade docs should appear in the search results.

docs/introduction.md
DavidBadura: I don't know if that's necessary. I think we will most likely only be creating pages, especially pages of this type.

PR #876, opened 2026-07-17
Remove unused public function `connection` on stores

DanielBadura:


DavidBadura (commented):
I would like to keep the methods because I find them useful.

DanielBadura:
Ok, I'll then open another PR which adds an interface so that you can better use these methods

PR #877, opened 2026-07-17
Add many missing Unit-Tests

DanielBadura:
This uncovered 2 issues:
* `SubscriptionRemoveCommand` was missing the parent call on configure (~needs a fix in 3.x too~ https://github.com/patchlevel/event-sourcing/pull/879)
* `AttributeEventTagExtractor` was converting numeric strings into `int` by using `array_keys`, we need to cast it to `string` again.


DanielBadura:
rebased, as connection functions will be kept, added one more test case covering it

DavidBadura (approved):

PR #879, opened 2026-07-22
Add missing parent call on `configure` for `Update SubscriptionRemoveCommand`

armetiz:
Add missing configure from parent.


DanielBadura:
Thank you! Could you run the cs fix?

